PR #125, opened 2015-05-02
Build

kshefchek:
Added build script


kshefchek:
Fails when no conf.json file is present, will fix.

PR #137, opened 2015-05-19
Clinvar fix

kshefchek:


kshefchek:
Build failed at panther db, orthologs_HCOP.tar.gz should be Orthologs_HCOP.tar.gz, will fix on this branch and remerge

PR #138, opened 2015-05-19
Clinvar fix round two

kshefchek:


kshefchek:
More panther issues, but going to merge as ClinVar is fixed now.  Will break out Panther into new issue.

PR #182, opened 2015-08-11
Update README.md

jmcmurry:
Parenthetical clarification for pip3 dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ like [Protege](http://protege.stanford.edu/).
 the code to skip this, if necessary
 
 * Required external python packages:
-    * [psycopg2](http://initd.org/psycopg/)
+    * [psycopg2](http://initd.org/psycopg/) 
     * [rdflib](https://code.google.com/p/rdflib/)
     * isodate
     * roman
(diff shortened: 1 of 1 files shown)


cmungall:
Any idea why travis is failing? I assume it's not sue to README changes!

kshefchek:
It's failing with nose.proxy.FileNotFoundError: [Errno 2] No such file or directory: 'owltools'

We need to either add the downloading/building of owltools into the dipper build or remove this test for now.

nlwashington:
ah, right, i think turned this test back on; i forgot why we were skipping it.  i really hate that we are skipping the owltools run in the standard testing suite.   what are our options here?

kshefchek:
If we know the commands to fetch and build owltools it would just be a matter of adding this to a "before_script:" tag in the travis yaml file, see https://github.com/monarch-initiative/monarch-app/blob/master/.travis.yml#L7

kshefchek:
I've added cloning/building owltools to the travis build, although it seems the only way for it to pass if to skip the maven tests.  I'll take a closer look at why it's failing.  But for now this should be sufficient.

cmungall:
@hdietze to advise

This will become a common request as many ontologies start using travis in their repos

Issue #278, opened 2016-03-10
Refactor code to use new OBAN property


mbrush:
We have been using OBAN:association_has_object_property  to link Monarch associations to its predicate. But this is not the intended use of this property.  OBAN has now added a new property specifically to link associations to predicates they use: OBAN:association_has_predicate.

All DIPper code needs to be refactored to replace all usage of OBAN:association_has_object_property with OBAN:association_has_predicate.  Should be straightforward as there should be no valid use of OBAN:association_has_object_property.

Full IRIs for these are:
http://purl.org/oban/association_has_object_property
to be replaced with
http://purl.org/oban/association_has_predicate

kshefchek:
Bump, we should fix this in the next release.

PR #280, opened 2016-03-14
  #278

TomConlin:
change 
association_has_object_property 
to 
association_has_predicate 

per #278


kshefchek:
This has implications downstream, lets make sure we can change this globally in all ttl files before merging in.

TomConlin:
I have undone the original small edits this pull release was for. They will be rolled into some other process that shares the cost of testing

kshefchek:
+1 merge away

PR #300, opened 2016-05-02
Use UniProt purl's for UniProt data

JervenBolleman:
The UniProt purls resolve to www.uniprot.org if html is requested but if the user wants RDF they get RDF. Using UniProt purls makes cross querying between sparql.uniprot.org and other data sources easier.


jmcmurry:
Excellent point, thanks @JervenBolleman

PR #312, opened 2016-05-17
change equivalent/sameAs to xref when processing dbxref table

kshefchek:
Changes property from equivalentClass/sameAs to OIO:hasDbXref, as this was causing equivalences between fly and human genes.


cmungall:
+1

jmcmurry:
Great fix, Kent. Thanks. Adding this scenario to [evidence of identifier pain](https://github.com/prefixcommons/operations/blob/master/evidence-of-identifier-pain.md) ...

PR #313, opened 2016-05-18
Ingest evidence and provenance from IMPC

kshefchek:
Changes:

**Source.py make_id()**
Add prefix argument, default is MONARCH

**GraphUtils.py _add_node()**
Handle blank nodes with the format _:1234 (previously only accepted _1234
Handle iris (previously just curies)
replace one letter variable names, python reserved names (id, object, etc)

**IMPC.py**
_add_evidence() - add evidence graph
_add_study_provenance() - add provenance graph

**Models/**
Evidence.py - functions for building evidence model
Provenance.py - functions for building provenance model

**resources/**
README.md
impc_mappings.yaml - mappings for various codes to iris (manual)
impc_parameters.yaml - web scraped code to iri mappings

**scripts/**
README.md
collapse-procedures.py
scrape-impc - scrapes impc to get parameter code to iri mappings

**tests/test_impc.py**
EvidenceProvenanceTestCase - tests output graph when passing test data into _add_evidence() and _add_study_provenance()

Anyone have bandwidth for a code review? @jmcmurry @TomConlin

To run tests:
python -m unittest tests.test_impc.EvidenceProvenanceTestCase


TomConlin:
stripping www. from the monarchinitative.org curies would be good
there is something wonky in that rapper (redland rdflib wrapper) chokes converting the output .ttl file to n-triples. but the same is true for the previous version. So perhaps not a blocker (but I still want to figure out why when there is time)
Will need to check but some ingests may import every curie as a @prefix needed or not and we have just doubled the number of curies.

kshefchek:
stripping www. from the monarchinitative.org curies would be good -> done!

kshefchek:
- Fixed IMPC models
- added initial reactome fetch/parse, functional test of parser/rdflib
- added tests for assertion model in impc
- impc test that fetches data and parses single row from file, and tests output graph
- script to generate triples of property declarations from list of ontologies (for loading test files into protege).

PR #328, opened 2016-06-10
Fixed bug in XSD:boolean URI

balhoff:
Boolean in dipper output currently look like `<http://www.ncbi.nlm.nih.gov/pubmed/11566854> :MONARCH_cliqueLeader "true"^^xsd: .` The python code was calling `XSD[bool]`, which works for some reason and just returns the namespace URI. Changed to `XSD['boolean']` so that the correct URI will be output as `xsd:boolean`.


balhoff:
I have not tested this change since I don't really have a dipper setup. And I am not sure if there might be any downstream code that would be checking for the incorrect datatype.

kshefchek:
cc @jnguyenx

jnguyenx:
Looks good to me.

Issue #342, opened 2016-06-28
Insist p-value scores are numeric


TomConlin:
animalqtldb  is interpreting  dbSBP rs_identifiers as p-values
a necessary  fix is to insist p-values are numbers only

A more through job would be to prevent them from being considered in the first place.

PR #343, opened 2016-06-28
AnimalQTLdb.py  to pass Jenkins tests

TomConlin:
ingest was finding rsid where it thought it had p-value score
made ignore all non numeric p-value scores 
(which does not address how or where the wrong  thing happens)

also at least one gene name is given where the id/symbol should be
fixed the particular case and preventing accepting spaces in future gene id/symbols

### DIFF
--- a/dipper/sources/AnimalQTLdb.py
+++ b/dipper/sources/AnimalQTLdb.py
@@ -1,3 +1,7 @@
+'''
+    TODO document me
+'''
+
 import csv
 import logging
 import re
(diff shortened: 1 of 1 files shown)


TomConlin:
address  #342

kshefchek:
Did they change the order of the columns?

TomConlin:
no, the code matches what I see (casually looking) at the data

TomConlin:
there may be a bad row somewhere but I did not dig for it

PR #349, opened 2016-06-30
refactor ensembl url from www to uswest

TomConlin:
seeing redirect problems from OHSU network (but not from Eugene)


jmcmurry:
This is fine, but if you configure to go to identifiers.org, it will direct to the geographically appropriate mirror for the data.

TomConlin:
the problem was in the redirect in the python library we are using

Issue #350, opened 2016-07-01
Update Clinvar for scigraph compatibility


kshefchek:
We need to make some updates to the clinvar xml parser to be compatible with some scigraph features.

1. Add the triple:
@prefix MonarchData: <http://data.monarchinitiative.org/ttl/> .
MonarchData:clinvar.ttl a owl:Ontology ;
Reason: This is used in SciGraph to link data to their source ontology

2. Use consistent RDFS:label iri
All rdfs label iris should be: http://www.w3.org/2000/01/rdf-schema#label
Right now some have a 1999 instead of 2000
In Scigraph, used to map to node.label property:
https://github.com/monarch-initiative/SciGraph-docker-monarch-data/blob/master/src/main/resources/monarchLoadConfiguration.yaml.tmpl#L109

3. Use blank nodes instead of monarch iris in the full set.  Blank nodes are handled differently in scigraph and UI.

Overall, it would be great if we were using RDFLib here, and our graph util and models API.  If there are pain points in using either of these please document these and we can figure out a gameplan to improve them moving forward.

TomConlin:
points addressed

PR #352, opened 2016-07-13
unexpand blank nodes

TomConlin:
address points in #350
add doc
output that the file is an ontology
make all types  rdf:type
allow unexpanded blank nodes
add script to partition clinvar xml test sets from full set and a list of RCVs


kshefchek:
+1 looks good to me.

I'm not sure if you want to add this to this PR, but it would be great if we could get rid of the duplicate curie configuration, which could lead to different curie mappings if one is changed without you updating the clinvar map, see https://github.com/monarch-initiative/dipper/blob/master/dipper/curie_map.yaml#L2 as an example.

kshefchek:
RDFLib has a rdf to dot conversion script as well, it may be useful to test against:
http://rdflib.readthedocs.io/en/stable/_modules/rdflib/tools/rdf2dot.html

TomConlin:
The output of the ntriple2dot script against the ingests for the current "GSA' release are here: 
http://data.monarchinitiative.org/dot/

PR #357, opened 2016-07-27
remove omim disease eqs from ncbigene

kshefchek:
Small fix to remove all omim disease eqs coming from ncbigene.


cmungall:
I assume we retain the equivs to OMIM true gene IDs?

kshefchek:
We do, the above approach checks scigraph-ontology to see if the omim ID is a gene.  It's probably not the best approach in retrospect, getting the information directly from MONDO would be better.

Issue #379, opened 2016-09-23
GENO_0000532 ! has expression-variant part


TomConlin:
@mbrush 
zfin ingest uses it but it is no longer found in https://raw.githubusercontent.com/monarch-initiative/GENO-ontology/develop/src/ontology/geno.owl

cmungall:
a quick grep around GENO reveals that this is the intended property: 
http://purl.obolibrary.org/obo/GENO_0000382

The fix should be straightforward but you will need to coordinate changes with any cypher queries

mbrush:
@TomConlin you can go ahead and swap in the GENO:0000382 property for any usage of GENO:0000532 in zfin.  I don't see any other ttl files with this predicate, and I don't see anything in the cypher queries that would need to be coordinated for this change.  Ticket can be closed once this is done.

Issue #377, opened 2016-09-23
GENO_0000208 ! has disposition


TomConlin:
@mbrush 
Our hpoa and  omia ingests use GENO_0000208 ! has disposition
but that term no longer appears in geno.owl

Will the term reappear or can you give me a replacement term?

mbrush:
This was a temporary relation in early versions of geno, which has since been replaced with http://purl.obolibrary.org/obo/RO_0000091. It can be replaced with this for now, and this issue closed.

However, the usage of this property in HPOA and OMIA is to link traits/phenotypes to their mode  of inheritance.  As per https://github.com/monarch-initiative/GENO-ontology/issues/22, the modeling for mode of inheritance patterns is under review and will be likely be updated in the near future.  

I have created a new ticket (#383) as  reminder to refactor the HPOA and OMIA sources once the inheritance modeling pattern is finalized.

Issue #386, opened 2016-10-27
Update gene label to id mapping to use new monarch search


kshefchek:
The monarch search has been updated and along with it broke our code that resolves gene labels to entrez ids.  This was caught by the UDP tests.

kshefchek:
Fixed with https://github.com/monarch-initiative/dipper/pull/418

PR #389, opened 2016-11-15
towards a new release

TomConlin:
this PR  is mostly comments but a few meaningful changes.
1) changes to ontology terms. (I have checked, and the correct terms are generated) 
   #377 #379
2) switch digest function from md5 to sha  for hash collision resilience 
   this would only matter if generated ID consistency between releases mattered   
   and if it did, switching sooner is better.    
3) animalqtl parse fix for new data decorations on 'range'
4) have ontology fetcher retry on the first timeout (helps with spurious ERO failures)


TomConlin:
@kshefchek 
Are there parts of the udp nosetests which need to be pushed?

kshefchek:
+1 looks good to me, feel free to comment out the udp test for now, see https://github.com/monarch-initiative/dipper/issues/386

Issue #385, opened 2016-10-18
ZFIN pheno_environment_fish.txt 


TomConlin:
@mbrush 
around line 2090 of ZFIN.py ingest we parse a file which in the function's comments refers to the non existent ```pheno_environment.txt``` but actually points to  <https://zfin.org/downloads/pheno_environment_fish.txt>
 which we are failing to parse as:   
```(environment_num, condition_group, condition, description, blank) = row```

because this ```pheno_environment_fish.txt``` file headers now includes  

```
Environment ID
ZECO Term Name
ZECO Term ID (ZECO:ID)
Chebi Term Name
Chebi Term ID (Chebi:ID)
ZFA Term Name
ZFA Term ID (ZFA:ID)
Altered Structure Name
Altered Structure ID (ZFA:ID)
NCBI Taxon Name
NCBI Taxon ID (NCBI Taxon:ID)
```
which is a different enough representation that it should get remodeled first.

mbrush:
Nice catch Tom - it does seem that ZFIN replaced its environment file with a new tsv with a very different format.  It leverages terms from its new ZECO ontology that they have been building.  We will need to update model and parser accordingly. Here we will want to try and follow patterns from [environmental conditions ontology](https://github.com/cmungall/environmental-conditions) that @cmungall is developing.
 
A couple options here:
 
1. One is to create instance level models from the data  in the [pheno_environment_fish.txt](https://zfin.org/downloads/pheno_environment_fish.txt) file that follow ECTO design patterns as best possible.  This should be fairly simple using their tsv file. 

  - **Example 1**:  for ZDB-EXP-150406-8 (ZECO:0000111!chemical treatment with CHEBI:16130!DDT) we would create a ZDB-EXP-150406-8 individual representing an instance of ZECO:0000111 and link this individual to the punned CHEBI:16130!DDT class IRI using a RO:0002233!has_input property.
 
[...]

PR #394, opened 2016-11-26
no zfin pheno env this release

TomConlin:
see #385


TomConlin:
tests fail because berkeleybop.org is down

Issue #380, opened 2016-10-04
@base_ landing page


TomConlin:
In our weekly dipper calls I have been agitating for a landing page for the vast number of unresolvable IRI we generate. (Where vast is defined as > 10,000,000 RDF statements)
So far we haven't come up a reason to keep it as is.  

Highest value source of these unresolvable IRI are __```@base``` ```underscore```__ nodes
which we seem to use for terms we wish had proper URI.

In curie form they are expressed as __```:_foo```__  which expands (or should expand to):  
 <https://monarchinitiave.org/_foo>.

There is no technical RDF reason we place a underscore after the colon
(and sometimes we don't).  

When the monarch web server has no idea what "_foo" is  (always?) it returns 404s.

Ideally these would go somewhere more useful, but if they cannot, 
I propose we have a destination page that acknowledges their existence.

There is a range of options between offering the same generic static page for everything  
and dynamic pages which include associated label, type, parent/children etc.  

Something along the lines of:

```
    https://monarchinitiative.org/iri/foo

                    foo
                    
    As far as we know this term "foo" does not have its own
    persistent resolvable url so we are using this one for now. 
    If you are aware of a another resource page for “foo” 
    please let us know us at  a@b.c

    This "foo" has the label "bar" and type of "fred".
    "foo" is linked to from:
        "something(s)"
    and links to:
        "otherthing(s)"
    
```     

In this scenario we leave the __@base__ prefix "__:__" alone as it
has other legitimate uses.  

Choose a specific @prefix e.g. __IRI__ : ```https://monarchinitiative.org/iri/```   
Which we use when we need a quick and dirty IRI for a wayward term.   

Make a landing page to capture anything under ```/iri/``` that tries to be as helpful as possible.

I have not checked, but I expect almost all of these @base_ terms will be leaf nodes in our graph  where the only reason for them not to be literals is the desire to attach properties.  

Being able to include the ingest source and date would be awesome.

cmungall:
Why can't we simply make these resolvable?

if dipper extracts it, it will be in scigraph-data. If it's in there we can query and find it's neighbors and categories and labels, and display a generic node landing page (not worry too much about making this look great, as you say goal is to not 404)

TomConlin:
That is pretty much exactly what I am requesting. I am not particularly attached to how it happens, the details above are just a suggestion to give something to chew on.

TomConlin:
For background here are the most frequently used base iri   
```
head base_iri.dist
[...]

PR #398, opened 2016-12-06
Flybase base nodes and breaking eutils

### LINKED ISSUE #397: Flybase's  ncbi eutil abuse
I had better look at our calls to eutils
running the flybase ingest  resulted  in:

```
    NCBI

Error
Access Denied

Your access to the NCBI website at www.ncbi.nlm.nih.gov has been temporarily blocked due to a possible misuse/abuse situation involving your site. This is not an indication of a security issue such as a virus or attack. It could be something as simple as a run away script or learning how to better use E-utilities, http://www.ncbi.nlm.nih.gov/books/NBK25497/, for more efficient work such that your work does not impact the ability of other researchers to also use our site. To restore access and understand how to better interact with our site to avoid this in the future, please have your system administrator contact info@ncbi.nlm.nih.gov.


Error
    blocked for possible abuse 
Server
    misuse.ncbi.nlm.nih.gov
Client
    76.115.148.221
Time
    Sunday, 04-Dec-2016 05:45:53 EST 
```

TomConlin:
takes first (big) steps on #380
closes #397 (eutils misuse)

### DIFF
--- a/dipper/curie_map.yaml
+++ b/dipper/curie_map.yaml
@@ -1,3 +1,7 @@
+# Skolemize Blank Nodes
+# overwrite iri fragment with '' for unresovable bnodes
+'_':    'https://monarchinitiative.org/.well-known/genid/'
+
 # Monarch-specific
 '': 'https://monarchinitiative.org/'
 'MONARCH': 'https://monarchinitiative.org/MONARCH_'
--- a/dipper/sources/FlyBase.py
+++ b/dipper/sources/FlyBase.py
@@ -3,6 +3,7 @@ import re
 import csv
 import gzip
 import io
+import hashlib
 
 from dipper.sources.PostgreSQLSource import PostgreSQLSource
 from dipper.models.assoc.Association import Assoc
@@ -1598,7 +1599,7 @@ class FlyBase(PostgreSQLSource):
                     if gene_id is not None and gene_id in self.label_hash:
                         # TODO FAIL: KeyError: None   default?
                         logger.info("getting label for gene_id:\t%s", gene_id)
-                        gene_label = self.label_hash[gene_id]  
+                        gene_label = self.label_hash[gene_id]
                     else:
                         if gene_id is None:
                             logger.error(
@@ -2077,19 +2078,24 @@ class FlyBase(PostgreSQLSource):
         """
         This is a special Flybase-to-MONARCH-ism.
         Flybase tables have unique keys that we use here, but don't want to
-        necessarily re-distribute those internal identifiers.
+        re-distribute those internal identifiers.
         Therefore, we make them into keys in a consistent way here.
+
         :param prefix: the object type to prefix the key with,
                     since the numbers themselves are not unique across tables
+                    nor guarenteed stable within a table over time
+
+        TEC: blank nodes created that way would not allways be valid RDF
+        Better is to use the input as a rdfs:label & maybe rdf:type
+        then digest to poduce a known valid key string.
+        Also other rdf tools DO rewrite blank node identifiers
+
         :param key: the number (unique key)
         :return:
 
         """
-        iid = '_fb'+prefix+'key'+key
-        if self.nobnodes:
-            iid = ':' + iid
 
-        return iid
+        return '_:' + hashlib.sha1('fb'+prefix+'key'+key).hexdigest()[1:20]
 
     def getTestSuite(self):
         import unittest
--- a/dipper/sources/FlyBase.py
+++ b/dipper/sources/FlyBase.py
@@ -1839,7 +1839,7 @@ class FlyBase(PostgreSQLSource):
                 # so we look them up using NCBI eutils services
                 tax_label = self.label_hash[organism_id]
                 # FIXME comment this out to speed things up
-                tax_num = DipperUtil.get_ncbi_taxon_num_by_label(tax_label)
+                # tax_num = DipperUtil.get_ncbi_taxon_num_by_label(tax_label)
                 if tax_num is not None:
                     organism_id = ':'.join(('NCBITaxon', tax_num))
                     self.idhash['organism'][organism_key] = organism_id
--- a/dipper/sources/FlyBase.py
+++ b/dipper/sources/FlyBase.py
@@ -2095,7 +2095,8 @@ class FlyBase(PostgreSQLSource):
 
         """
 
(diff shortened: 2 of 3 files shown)


dipper/curie_map.yaml
```diff
@@ -1,3 +1,7 @@
+# Skolemize Blank Nodes
+# overwrite iri fragment with '' for unresovable bnodes
+'_':    'https://monarchinitiative.org/.well-known/genid/'
```
kshefchek: I assume this will change our bnodes globally, which is fine, but we need to make sure that these can still get labelled as anonymous in SciGraph.  We currently use the OwlAPI - http://owlapi.sourceforge.net/javadoc/org/semanticweb/owlapi/model/OWLIndividual.html#isAnonymous() method in particular -  to assign anonymous labels to nodes in Neo4J, and these labels are used elsewhere (cliqueLeader processing, generation of solr association views).

I'm not sure the best way to go about this, is there an explicit type "anonymous" in either the rdf or owl specs?  @cmungall @mbrush curious what your thoughts are here.

TomConlin: Just looking at this in RDFLIB

http://rdflib.readthedocs.io/en/stable/_modules/rdflib/term.html#BNode

     BNode.skolemize(authority='http://rdlib.net/')

        Create a URIRef “skolem” representation of the BNode, in accordance with
        http://www.w3.org/TR/rdf11-concepts/#section-skolemization

and 

http://rdflib.readthedocs.io/en/stable/apidocs/rdflib.html#rdflib.term.URIRef.de_skolemize

kshefchek: This makes sense but I'm unclear if the OWL API will correctly interpret these as anonymous.

Note that it's possible that removing the anonymous distinction may not have much of a negative effect downstream.  Right now I see it primarily being used to query out variants and setting clique leaders.  A lot of this could be fixed by re-modelling our g2p and g2d associations, as suggested in #380.

TomConlin: interpreting well defined rdf blank nodes 
as neo4j anonymous nodes seems the only sensible thing to do.  
If we are relying on the first character of a post curi-prefix/iri-fragment identifier 
to be an underscore to determine if a scigraph node is anonymous, then that needs addressing.    
If neo4j/scigraph are already correctly interpreting standard blank nodes then nothing would break until we actually emitted skolems. 
Anonymous nodes should most likely already be getting generated for any iri   
with the pattern "_host_ /.well-known/genid/"  as they could already be occurring in the wild.

kshefchek: I may be misunderstanding how the pieces fit together, the easiest thing is probably to run a small file through the scigraph OWL Loader and evaluate the output.

TomConlin: already done on my local repo

kshefchek: Or another (maybe not so great) option:
node = node.skolemize(self.curie_map.get(''))
node = re.sub(r'rdflib\/', '', node)
node = URIRef(node)

kshefchek: I did a test load and can confirm that the OWL API doesn't appear to recognize these as anonymous.

        Anonymous nodes should most likely already be getting generated for any iri
        with the pattern "host /.well-known/genid/" as they could already be occurring in the wild.

Perhaps this is something we can do in SciGraph, @jnguyenx?

TomConlin:
> @kshefchek I assume this will change our bnodes globally, which is fine, but we need to make sure that these can still get labeled as anonymous in SciGraph.  

Yes and no. it will only change future behavior.
Just want to make it explicit, we have a full set of files ready for release which are not subject to this change in bnode behavior. For this release, this change is only to be applied to the flybase nodes which are currently broken anyway. We will then have till the next release to iron out correct scigraph/skolem behavior.

kshefchek:
Understood, +1 on merging in.

dipper/curie_map.yaml
jnguyenx: We can add a post-processor in SciGraph with a config saying: Tag nodes with these prefixes as anonymous. I think that'll do the trick.

cmungall: This really needs to be well documented. In RDF the concept of a blank node is unambiguous. Once you skolemize into a URI, it's not anonymous, period.

I see the use case for being able to tag used-to-be-blank nodes, but I don't think we should confuse terminology. We should call these pseudo-anonymous or something. And we need something on the main scigraph docs about them (or if this is something purely localized to monarch convention, we still need this documented somewhere).

I also think encoding a secret convention between scigraph and dipper about URI syntax is a bit fragile. Probably best to keep right now, but really this should be more explicit. E.g. if dipper wants to mark a URI in some way, then it should mark is explicitly (with a triple), not by convention

jnguyenx: @cmungall is there a way in owl to declare a node as anonymous?

The owlapi has this function: 
```
public interface OWLIndividual extends OWLObject, OWLPropertyAssertionObject {
...
    /**
     * Determines if this object is an instance of
     * {@link org.semanticweb.owlapi.model.OWLAnonymousIndividual} Note that
     * this method is the dual of {@link #isNamed()}.
     * 
     * @return {@code true} if this object represents an anonymous individual (
     *         {@code OWLAnonymousIndividual)} or {@code false} if this object
     *         represents a named individual ( {@code OWLIndividual})
     */
    boolean isAnonymous();
```

cmungall: > is there a way in owl to declare a node as anonymous

Not as such. If your axiom is about an IRI, it's about an IRI. If it's about a blank node, it's about a blank node. 

Of course we can create our own annotation property called say 'pseudo-anonymous', and use that to annotate any IRI.

kshefchek: We really just need a class for IRIs that are named, but otherwise follow the same conventions/use cases as RDF blank nodes (ie describe reification, multicomponent structures, etc) and are not guaranteed to persist across releases.

We could create a class in the monarch data Monarch:Pseudonymous and use this to label nodes in SciGraph (as either anonymous, or pseudonymous)?

TomConlin: +1  dipper produces RDF but should strongly limit its notion of blank nodes to be no more than OWL allows. 
<https://en.wikipedia.org/wiki/Blank_node#Anonymous_classes_in_OWL> 

    anonymous classes such as unions or intersections of classes, or classes called restrictions, defined by a constraint on a property.

PR #405, opened 2017-01-07
Catch SAX parser exception, other fixes

### LINKED ISSUE #401: Add try except to handle variable .owl formats
OWL file formats appear to be variable (is this correct @cmungall), see http://purl.obolibrary.org/obo/pco.owl as an example.  Instead of expecting XML we should add try/except clauses to test various acceptable formats.

Note this was caught by a travis build: https://travis-ci.org/monarch-initiative/dipper/builds/187602697#L666

kshefchek:
- Catch sax errors for owl files formatted as ttl
- Remove some cache files from github
- Clean up IMPC impress code map generation
- Fix other IMPC parser bugs

closes #401


TomConlin:
+1   
running on GO does not result in sax error,   
impc completes and is slightly larger than previous run

Issue #147, opened 2015-06-10
BGEE: Write a dipper importer for gene expression data


cmungall:
bgeedb have done all the work here, no need to go to individual MODs, http://bgee.org/

@fbastian -- this would be to create turtle files from the bgee downloads (with full attribution of course). These could then be used to make monarch anatomy pages, analogous to http://beta.monarchinitiative.org/phenotype/HP:0000315 -- not sure what yr current plans are for a UI here, maybe we could join forces here?

fbastian:
We don't currently have plans for an "anatomy" page. But of course we would be glad to join our efforts, and that could be the opportunity for us to develop such a page. Contact me whenever you want.

mellybelly:
I would be so incredibly delighted to have you help us with anatomy pages! with ontology navigation, pictures, homologies, ooohhh yah.

cmungall:
bgee provides expression values and expression calls: http://bgee.org/?page=download

We would use the already pre-processed calls

http://bgee.org/?page=download&action=expr_calls#id1

options are:

 * Presence/Absence of expression
 * Over-/Under-expression across anatomy
 * Over-/Under-expression across life stages

I suggest leaving stages til a second iteration. Of the first two, we would seek guidance. As a first pass we would not use this algorithmically in owlsim (that will take further consideration, and we'd probably want to include other sources). This would simply be a list of tissues on a gene page, and a way of providing a good uberon page that shows relevant genes.

Either way we'd not do our own processing. We'd take the bgeedb tsvs and make a simple OBAN type model of each row, using RO:0002206](http://purl.obolibrary.org/obo/RO_0002206) ! expressed in

mellybelly:
and include the negation in first iteration too? I think that is important

kshefchek:
Do we want to filter out poor call quality?

mellybelly:
I would think so, at least in first pass. alternatively could put a UI flag that highlights that fact.

mbrush:
For the basic expression data, seems like each oban:association would represent a ternary relationship between a gene, anatomy, and stage.  The association would link the gene (**subject**) and anatomy (**object**) using RO:0002206 ! expressed in (**predicate**) - with the stage as a qualifier hanging from the association (analogous to G2P associations where stage is a qualifier on the association).  Looks like there is some nice evidence metadata, particularly for the differential expression data (p-values!). Any objections to bringing this in on first pass?  

A couple questions:

[...]

PR #407, opened 2017-01-13
BGee ingest

kshefchek:
First pass at Bgee ingest as described in https://github.com/monarch-initiative/dipper/issues/147


kshefchek:
First pass Bgee fetcher/parser implemented, will write some tests before merging in.

PR #414, opened 2017-01-22
MyDrug Aeolus subset data ingest

kshefchek:
Ingest of Aeolus data from MyDrug.


kshefchek:
First pass (about half the model)
[mydrug.txt](https://github.com/monarch-initiative/dipper/files/721714/mydrug.txt)

PR #416, opened 2017-02-01
Refactor graph_utils, models API, bnode generation

kshefchek:
This is a global refactor that cleans up our models API,

- define abstract base class for a graph, break out rdflib and streaming graph implementations
- split graph_utils into three classes, models class containing basic rdf/owl axioms (addLabel, addIndividual, addType, etc)
- remove nobnode code, bnode skolemization moved to graph class


cmungall:
+1

some things for later

 - is convenient to just write `dct:isVersionOf` but should we think of a vocabulary module for stronger typing and better introspection
 - this might also be able to take care of some of the repetition of some of the prefix info

cmungall:
is this a false positive fail in travis?

kshefchek:
it's a true positive, still working through updating each source.

kshefchek:
@cmungall @TomConlin this is pretty much ready to go, let me know if you have any comments.

Issue #420, opened 2017-02-11
ClinVar: lost gene-variant mappings


TomConlin:
continued from  https://github.com/monarch-initiative/monarch-app/issues/1398

mbrush:
@TomConlin  the list of variant categories you pulled back using

     xmlstarlet sel -t -v './ReleaseSet/ClinVarSet/ReferenceClinVarAssertion/MeasureSet/Measure/@Type' ClinVarFullRelease_00-latest.xml.gz  | sort | uniq -c | sort -nr

may not be querying the right place to return the categories we need, as the list seems to be missing 'haplotype' which was the category that motivated us to look at this in the first place.  Could there be another element in the xml that we need to hit?  or perhaps this set of categories is encoded outside the xml data itself?

TomConlin:
the other logical place to look is the same field within the SCVs

    $ xmlstarlet sel -t -v './ReleaseSet/ClinVarSet/ClinVarAssertion/MeasureSet/Measure/@Type' ClinVarFullRelease_00-latest.xml.gz  | sort | uniq -c | sort -nr
     355876 Variation
      30305 single nucleotide variant
       7936 copy number gain
       7294 copy number loss
       2398 Deletion
        965 Duplication
        251 Translocation
        217 Indel
        186 Insertion
         38 Complex
          4 Tandem duplication
          1 Inversion

nope not there either.
How about up a level?  
  
    xmlstarlet sel -t -v './ReleaseSet/ClinVarSet/ClinVarAssertion/MeasureSet/@Type' ClinVarFullRelease_00-latest.xml.gz  | sort | uniq -c | sort -nr
     404920 Variant
        164 CompoundHeterozygote
         95 Haplotype
  
okay something there but not enough to say so.
how about up a level in the RCV?  
    xmlstarlet sel -t -v './ReleaseSet/ClinVarSet/ReferenceClinVarAssertion/MeasureSet/@Type' ClinVarFullRelease_00-latest.xml.gz  | sort | uniq -c | sort -nr
     365344 Variant
        258 Haplotype
        175 CompoundHeterozygote
         5 Phase unknown
          1 gene-variant
          1 Distinct chromosomes

 Well I guess that is about as good as it gets.

mbrush:
Interesting.  So it seem that we need to interrogate two elements in the RCV to determine how to rdf:type variants in ClinVar:
 
[...]

PR #421, opened 2017-02-11
ClinVar updates

TomConlin:
https://github.com/monarch-initiative/dipper/issues/420

Precipitated by an increase in granularity  for variant relation types 
but became a general rewrite of mapping behavior.

Here mapping has explicit distinct steps even when those steps are merged in practice
the first step is mapping from what is found in the external dataset to the preferred labels 
we use within the code.  The second mapping is from the preferred label used within the code to the ontological term identifier emitted as RDF.  The first translation table is particular to the ingest and may map multiple external strings to the same preferred label. The second associates preferred labels and their ontological terms and are to be shared by all ingests.  
Both mapping files may change when ontologies are updated.
The local translation table can need to be changed when the data changes 

other notes:  

ClinVar xml parser changed from DOM to (chunky) SAX style parsing because DOM parsing
was exceeding 32G (6 months ago it was ~24G)  the SAX parser is limited to the size of the output (less than 2G) plus a small constant for the current chunk which should be fine for the foreseeable future.  

The genotype and genomicfeature files are innocents caught in a merge.  
The .tt mapping files will probably end up as .yaml (although I like data ala cart) .  
Tiny tweaks to the dot scripts are not relevant.


kshefchek:
+1, make sure to switch has_object_property to has_predicate

TomConlin:
@kshefchek  
I am also noticing a dearth of disease labels which needs fixing

PR #423, opened 2017-02-23
more on #420 

TomConlin:
An eclectic collection

- changes to support archive.monarchinit ... existing
- move data-monarchinit...curie  to httpS 
- another iteration on supporting more granular variant types in clinvar 
- switch bnodes to skolemized by default in clinvar 
- switch association_has_object_property to association_has_predicate
- update sources in runall script in the test directory
- include a new script to isolate a subset of ClinVar as a test-set given various IDs

note the clinvar subset currently specified is ... too generous. Returns about a third of the release which is larger than most ingests and far too big for testing  but that will be another ticket.


kshefchek:
+1

TomConlin:
the failed check is for fetching panther which will will have nothing to do with this PR

TomConlin:
the fail is pulling panther data files which is not likely to be related to these chnges

PR #424, opened 2017-02-23
Bnode annotation

kshefchek:
Annotates variant blank nodes with MONARCH:anonymous annotation property.  Long term the model of anonymous variants to proxy G2P relationships will be removed.


TomConlin:
+1  
again with the panther fetch fail. they must be having a bad day

PR #460, opened 2017-05-19
ingest maintaince

TomConlin:
- zfin added evidence code to disease file, also un needed carriage return filter was broken 
- omim annual api key update.  runs to completion but noticed many elpty ojects being created . should investigate further
- clinvar have an email in to clinvar inquiring about the xml ""MeasureSet" irems which 
without warning no longer have an "ID" attribute. it will not be hard to adapt but suspect is is a mistake.  other than that i had been making the language around local & global translation tables more consistent 

note clinvar is still expected to fail

### DIFF
--- a/dipper/graph/RDFGraph.py
+++ b/dipper/graph/RDFGraph.py
@@ -42,7 +42,8 @@ class RDFGraph(ConjunctiveGraph, DipperGraph):
                     (self._getNode(subject_id), self._getNode(predicate_id),
                      Literal(obj)))
             else:
-                logger.warn("Null value passed as object")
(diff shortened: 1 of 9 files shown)


kshefchek:
Should we configure ClinVar to pull from an older version in the meantime?  What are your thoughts on combining file version, label term, and translation table into one single config file (would likely need some k-v format - yaml, json?)

kshefchek:
There is a 5/23 release of clinvar, the latest file lags behind their weekly releases:

ftp://ftp.ncbi.nlm.nih.gov/pub/clinvar/xml/weekly_release/ClinVarFullRelease_2017-0523.xml.gz

PR #471, opened 2017-05-30
mgi running and  host change for fly 

TomConlin:
- flybase moved their read only postgresql instance to a machine named `chado`
  - unannounced, found by looking for pages with recent edits in their wiki
- in MGI guard against stray tabs in TSV by skipping rows with the wrong number of columns
  - tabs are not invalid in the database and are not escaped when selected out
- we were producing blank nodes with double colons  `_::yadayamgibackground`

Have a ticket in to mgi to address the specific case we are working around
suggest they trim leading/trailing whitespace of freetext on ingest.

Flybase is not completing because a FBgnID worms (flutters?) into dbxref and we do not supply a curie for it.    
note Flybase is not sending the FBgn in their dbxref dataset we are doing this ourselves.

### DIFF
--- a/dipper/sources/FlyBase.py
+++ b/dipper/sources/FlyBase.py
@@ -164,7 +164,7 @@ class FlyBase(PostgreSQLSource):
 
         # create the connection details for Flybase
         cxn = {
-            'host': 'flybase.org', 'database': 'flybase', 'port': 5432,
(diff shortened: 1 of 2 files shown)


TomConlin:
per email reply

The stray tab has been removed. The updated file will be available next week for public.

Ref:MSG066950

PR #473, opened 2017-06-02
Allow bgee version to be passed via cmd line

kshefchek:
Allows version to be set on the command line for bgee and string.

Note - we should do this for other sources as well.


TomConlin:
+1

Issue #444, opened 2017-03-30
Use biomart API for up-front mapping and to simplify Ensembl module


cmungall:
Consider replacing

E.g.
https://github.com/monarch-initiative/dipper/blob/077ac9672c3cbd9af19a620fa2644725899a115f/dipper/sources/Ensembl.py#L134-L200

(lots of hardcoded species..)

with https://pypi.python.org/pypi/biomart/

We may also want to use this for sources like string, that use ENSP. While we could delay mapping to clique merge stage, this may be expensive and without much gain.

It may be simpler to up-front map. ie. the String.py module would use biomart.py to get ENSP->Gene for each species and map as part of ingest.

kshefchek:
Chatted with this last week with @mellybelly and @mbrush and they both felt strongly we leave the ENSP to ENSP relations instead of summing to the gene level at ingest time.

kshefchek:
I reread some docs today that point to String not mapping data at the protein level, but rather using a single isoform as a placeholder:

STRING uses one protein per gene. If there is more than one isoform per gene, we usually select the longest isoform, unless we have information that suggest that other isoform regarded as cannonical (e.g., proteins in the CCDS database).

From http://string-db.org/cgi/help.pl, FAQ

Perhaps we should reconsider an upfront mapping to genes? cc @mellybelly @mbrush

kshefchek:
Fixed with https://github.com/monarch-initiative/dipper/pull/478

PR #478, opened 2017-06-08
use try/except instead of searching large dict

kshefchek:
- Type uniprot proteins
- upfront map proteins to genes
- use try/except instead of searching keys in large dictionary

https://github.com/monarch-initiative/dipper/issues/444#issuecomment-306870845


kshefchek:
String may be rate limiting our downloads, although I'm not getting a 429 error.  This is the only way I can get it to run on monarch5:

wget -qc --tries=20 --read-timeout=2

Otherwise this script works fine on my local computer, and other ftp/http downloads over monarch5 are fine.  We can revisit if it is a ongoing issue.

PR #487, opened 2017-07-11
hopa and clinvar fixes

TomConlin:
clinvar has a new method type "phenotyping only" we will need a sepio term for 
(only several records have it now)
hopa  just needs more scrubbing. 
 - leading/trailing spaces 
 - spaces (expanded tab?) between curie and local identifier

a few flybase comments and start moving moving away from select *


kshefchek:
+1 thanks for the quick iteration
For ClinVar, we should get clarification on what "phenotyping only" means.  I find the term a little confusing.

I don't think it would be terrible to use SEPIO:0000081 ! curation based study, since that is what we use for G2P associations generated from some heuristic approach (ACMG for example).  But @mbrush might disagree.

TomConlin:
https://github.com/monarch-initiative/SEPIO-ontology/issues/17 is where I put the identifiers for the only known examples of  `phenotyping only`.

mbrush:
@kshefchek  - i don't think this is a curation-based study, as I assume it involves collecting phenotype data from actual patients.   I saw nothing about this term in the clinvar release notes, but did find some info on the ClinGen website [here](https://clinicalgenome.org/share-your-data/clinicians/clinvar-submission/). 

I will email Melissa Landrum at ClinVar to ask what exactly is meant, but I think it is safe to create a new term for this method.  So I added SEPIO_0000186 ('patient phenotyping study') to support these records.

TomConlin:
have added the  sepio term and mapping and removed the filter

PR #493, opened 2017-07-19
Create incoming_curie_map.yaml

jmcmurry:
This is a woefully incomplete start at what classes and sources of entities are served as pages within the monarch app. An attempt to see if this kind of a file would be useful to address https://github.com/monarch-initiative/monarch-app/issues/1201. Broadly speaking, I would like to see this fleshed out for knowledge beacon purposes and get Monarch integrated as a 3rd party provider within n2t/identifiers.org registry.


cmungall:
We should talk about this. Need to make it clear the monarch URIs have some kind of isAbout relationship to source

TomConlin:
will need to evolve a way of maintaining this file

PR #508, opened 2017-08-22
Created RGD ingest script  and added 'add_date_created()' module to Provenance.py

putmantime:


putmantime:
I need to add more documentation before this is merged

Issue #509, opened 2017-08-23
Clinvar variants not associated with genes


TomConlin:
ClinVar now has variants not associated with a gene for example [RCV000051577](https://www.ncbi.nlm.nih.gov/clinvar/RCV000051577/)  

hence it will have have no gene symbol which is a new behavior

others  
```
RCV000050883 VARIANT MISSING Gene Symbol
RCV000051577 VARIANT MISSING Gene Symbol
RCV000051723 VARIANT MISSING Gene Symbol
RCV000052394 VARIANT MISSING Gene Symbol
RCV000052395 VARIANT MISSING Gene Symbol
RCV000052925 VARIANT MISSING Gene Symbol
RCV000053269 VARIANT MISSING Gene Symbol
RCV000133780 VARIANT MISSING Gene Symbol
RCV000134139 VARIANT MISSING Gene Symbol
RCV000134174 VARIANT MISSING Gene Symbol
RCV000134948 VARIANT MISSING Gene Symbol
RCV000135843 VARIANT MISSING Gene Symbol
RCV000136687 VARIANT MISSING Gene Symbol
RCV000138139 VARIANT MISSING Gene Symbol
RCV000139166 VARIANT MISSING Gene Symbol
RCV000139435 VARIANT MISSING Gene Symbol
RCV000139806 VARIANT MISSING Gene Symbol
RCV000141813 VARIANT MISSING Gene Symbol
RCV000142008 VARIANT MISSING Gene Symbol
RCV000142245 VARIANT MISSING Gene Symbol
RCV000143710 VARIANT MISSING Gene Symbol
RCV000207023 VARIANT MISSING Gene Symbol
RCV000494707 VARIANT MISSING Gene Symbol
RCV000203445 VARIANT MISSING Gene Symbol
RCV000207053 VARIANT MISSING Gene Symbol
RCV000207313 VARIANT MISSING Gene Symbol
RCV000207326 VARIANT MISSING Gene Symbol
RCV000445663 VARIANT MISSING Gene Symbol
RCV000445925 VARIANT MISSING Gene Symbol
RCV000446345 VARIANT MISSING Gene Symbol
RCV000446555 VARIANT MISSING Gene Symbol
RCV000446684 VARIANT MISSING Gene Symbol
RCV000447056 VARIANT MISSING Gene Symbol
RCV000448130 VARIANT MISSING Gene Symbol
RCV000448186 VARIANT MISSING Gene Symbol
[...]

PR #510, opened 2017-08-24
Translation Table files to YAML and allow intergenic variants

TomConlin:
Tim is going to create translation tables as yaml so the existing ones need converting
Clinvar is now including variants which cannot have a gene symbol


TomConlin:
closes  #509  
note travis reports this passed more than 20 min ago
github reports travis is saying it is still pending

TomConlin:
closes  #509  
note travis reports this passed more than 20 min ago
github reports travis is saying it is still pending

PR #521, opened 2017-09-16
join cvterm and feature_relationship, filter on label instead of prim…

kshefchek:
Removes primary keys when processing feature_relationship, joins table with cvterm and filters on labels.  Note these are different from the obo cvterms.   I pulled the 2015 version of the flybase db to ensure the original key-label mappings are accurate.


kshefchek:
Clinvar bug caused by two entries in the translation table:
"sequence_alteration": "SO:0001059"
...
"sequence_alteration": "SO:000105"

Issue #525, opened 2017-09-26
Invalid values for datatype in animalqtldb_dataset.ttl


yy20716:
Hello, I tried to import animalqtldb_dataset.ttl file (https://data.monarchinitiative.org/ttl/animalqtldb_dataset.ttl) into a graph database such as Blazegraph. However, a number of errors occured as follows and the import process was failed.

> [error] (run-main-0) org.openrdf.rio.RDFParseException: '2017-09-25-09-00' is not a valid value for datatype http://www.w3.org/2001/XMLSchema#dateTime [line 14]
> org.openrdf.rio.RDFParseException: '2017-09-25-09-00' is not a valid value for datatype http://www.w3.org/2001/XMLSchema#dateTime [line 14]

It is possible that the time data type is not correctly expressed, e.g., 2017-09-25-09-00 may need to be represented as 2017-09-25T09:00:00. Could you please check this issue? Thank you.

kshefchek:
thanks for fixing!

PR #526, opened 2017-09-27
Modify the format of date_accessed value to conform xsd:dateTime format.

yy20716:
* This patch is made to solve the issue posted at https://github.com/monarch-initiative/dipper/issues/525. 

* I was testing Blazegraph triplestores with ttl files in the data repository (https://data.monarchinitiative.org/ttl) and found that Blazegraph's parser did not allow importing some *_dataset.ttl files because object values of dct:issued did not conform xsd:dateTime format, i.e. 2017-09-25-09-00 should be 2017-09-25T09:00:00. I tracked the codes and found that the value was generated by line 35 in Dataset.py (https://github.com/monarch-initiative/dipper/blob/master/dipper/models/Dataset.py). I fixed the format of the date to conform the xsd:dataTime format so that ttl filescan be imported without errors.

### DIFF
--- a/dipper/models/Dataset.py
+++ b/dipper/models/Dataset.py
@@ -32,7 +32,11 @@ class Dataset:
         self.identifier = ':' + identifier
         self.version = None
         self.date_issued = None
-        self.date_accessed = datetime.now().strftime('%Y-%m-%d-%H-%M')
+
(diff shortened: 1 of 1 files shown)


TomConlin:
going to guess you will need to get access to a private monarch config repo

PR #544, opened 2017-11-12
OMIA

TomConlin:
seemed to be an (intermittent?) problem following redirects

### DIFF
--- a/dipper/sources/OMIA.py
+++ b/dipper/sources/OMIA.py
@@ -54,7 +54,10 @@ class OMIA(Source):
     files = {
         'data': {
             'file': 'omia.xml.gz',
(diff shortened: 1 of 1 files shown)


TomConlin:
between local testing an running via jenkins
input changed.

now has a link to OMIM:10700 which does not exist 
have a email in suggesting they append a zero making it OMIM:107000
which is believable scanning the papers.

TomConlin:
OMIA  has replied they will try and fix it within a day or so

kshefchek:
Can you inquire about their drop in data? I never sent a second email.

TomConlin:
Already did by way of requesting release notes:

     ...am wondering if there could be some sort of OMIA release notes 
    to help me figure out why sometimes the data just seems to disappear.

PR #546, opened 2017-11-14
restore WB xref; output RDF formats; var keyword abatment; remote file change detection 

TomConlin:
- wormbase played hide the xref file
 - extend the RDF formats  data files may be output as
    - now includes nquads
 - ensure `_dataaet` and `_test` files are only written as turtle 
 - reduce the usage of python keywords as user variables i.e `format`
 - make http headers have RCF specified case (I believe lib took care of it anyway)
 - corrected string representation of number that likely tainted file size comparisons

### DIFF
--- a/dipper-etl.py
+++ b/dipper-etl.py
@@ -125,8 +125,8 @@ def main():
         action="store_true")
 
     parser.add_argument(
-        '--format',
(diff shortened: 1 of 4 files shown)


kshefchek:
+1

TomConlin:
jenkins jobs updated as well

PR #555, opened 2017-12-27
Fixes for flybase, hpoa, mmrrc

kshefchek:
- Removed primary key reference for cvterm, replaces with join and matching on cvterm label
- Adjustment for 15th column in hpoa, note the hpoa file format will be refactored in the new year
- update udp test with hgnc instead of entrez gene
-  Fix mmrrc whitespace issue

Test runs:
http://ci.monarchinitiative.org/job/build-mmrrc-ttl/40/
http://ci.monarchinitiative.org/job/build-hpoa-ttl/53/
http://ci.monarchinitiative.org/job/build-flybase-ttl/76/

### DIFF
--- a/dipper/sources/FlyBase.py
+++ b/dipper/sources/FlyBase.py
@@ -57,8 +57,7 @@ class FlyBase(PostgreSQLSource):
         'dbxref',               # done
         'phenotype_cvterm',     # done
         'phendesc',             # done
-        'environment_cvterm',   # done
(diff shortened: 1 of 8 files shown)


TomConlin:
+1

new file?  `scripts/treeify-hpo.py`

kshefchek:
It was some work for pcori, but probably belongs in a different repo, just storing here for now.

PR #557, opened 2018-01-18
integrate OMA as new orthology resource for monarch

alpae:
Hi,

this pull-requests contains a new source of orthology data based on orthoxml format in general and a concrete instance of OMA providing the data based on the Quest for Orthologs reference proteomes data.

Adrian


kshefchek:
Thanks for PR and new source! The tests will fail currently for unrelated reasons.  I can work on a fix so that we can merge this in.

dipper/sources/OrthoXML.py
```diff
@@ -0,0 +1,229 @@
...
+
+                # might as well add the taxon info for completeness
+                g.addTriple(
+                    protein_id_a, model.object_properties['in_taxon'], taxon_a)
+                g.addTriple(
+                    protein_id_b, model.object_properties['in_taxon'], taxon_b)
+
+                assoc.add_association_to_graph()
+
+                if not self.testMode \
```
kshefchek: We would like to move away from this method of testing.  If you would like to create a subset file, I recommend generating a smaller xml file and using that as input, or generating the RDF dataset and querying out a subset with sparql.

dipper/sources/OrthoXML.py
```diff
@@ -0,0 +1,229 @@
...
+
+                rel = self._map_orthology_code_to_RO[rel_type]
+                evidence_id = 'ECO:0000080'  # phylogenetic evidence
+
+                # add genes to graph;
+                # assume labels will be taken care of elsewhere
+                protein_id_a = self.clean_protein_id(protein_id_a)
+                protein_id_b = self.clean_protein_id(protein_id_b)
+                model.addClassToGraph(protein_id_a, None)
+                model.addClassToGraph(protein_id_b, None)
```
kshefchek: Can we type these as SO:0000104? 
i.e. model.addClassToGraph(protein_id_a, None, "SO:0000104")
or 
model.addClassToGraph(protein_id_a, None, Genotype.genoparts['polypeptide'])

@mbrush does it matter if these are individuals or classes?

kshefchek:
@alpae if you pull from master and push the tests should pass.  Everything looks good when I run it.  However, in some cases we may need to map the protein identifier to the gene identifer (via the predicate GENO:0000418).  For example, I don't think we currently link FlyBase protein identifiers to genes.  Ideally we would get this from flybase, but if OMA already stores this it may be easier to do this here for now.  Ensembl and UniProt protein to gene should be handled in other parsers.

dipper/sources/OrthoXML.py
alpae: sure. I started with what the Panter.py source was doing. is there already some simple example around that implements your prefered way of testing?

dipper/sources/OrthoXML.py
alpae: yes, seems like SO:0000104 is exactly what it is. will update accordingly. Thanks for pointing out

alpae:
@kshefchek not sure that FlyBase indeed is a source of the data. The idea we had with @selewis is to use the QuestForOrthologs reference proteomes dataset that is maintained by UniProt. The first release (which is what I was using as a proof of concept mainly) contained certainly proteins from Ensembl and UniProt, but very likely none from FlyBase. Will remove that namespace case for clarity if it is not used in any of the Reference Proteomes Datasets as primary ID.

kshefchek:
Okay great, for the taxon settings, will it retrieve the input taxa vs all organisms, or taxa vs taxa.  We may want to break these out as options.  Do you have a sense for how long it takes to run over a few species?

alpae:
Just realized that the taxon setting is not yet doing anything, but was intended to just filter relations that are limited to a set of taxa, e.g. orthologs only among a set of species. 
Processing the whole file takes quite long, about a day if I remember correctly.

alpae:
@kshefchek implemented the changes you asked for and did another run with the whole dataset. it takes something like 24hrs to process all the relations and their genes (it's ~60mio relations)

selewis:
Kent started a Jenkins job, so it's building now: http://ci.monarchinitiative.org/view/dipper/job/build-oma-ttl/1/

kshefchek:
I performed a test with taxa limiting with the cmd:
> python dipper-etl.py --sources oma --taxon 7955,9606,10090,6239,7227,10116,559292,4896,4932

However this outputted an empty file.  Is it possible that one of these taxa do not exist and it is failing siltently?

I performed another test with no taxa limiting but the run took a while and I needed to free up resources on the machine this was running on.  I'm running another test now: http://ci.monarchinitiative.org/view/dipper/job/build-oma-ttl/4/

alpae:
I've realized that in the data file we load there are no proper taxonomic ids stored. That's why if you filter for certain species, the result is empty. 

I'm currently anyways computing the OMA predictions on the newer QfO reference proteome dataset. I expect the new input file to be ready within a few days. There, the taxonimic ids should be properly recorded. If you wish to progress before, I can also do a quick hack to map the species ids to the taxonomic ids for the existing dataset...

alpae:
Hi @kshefchek, I've updated the underlying dataset to the QfO reference proteome dataset from 2017 (the latest one), and made sure the taxonomy information is properly stored. I locally tested with a limited set of taxons and this seems to work now properly.

selewis:
Kent started the load today (http://ci.monarchinitiative.org/view/dipper/job/build-oma-ttl/6/console). We'll can check in tomorrow to see what happens

kshefchek:
+1

Issue #437, opened 2017-03-10
Use central ECO GAF mapping for GO parser rather than hardcoding our own


cmungall:
Use this
http://purl.obolibrary.org/obo/eco/gaf-eco-mapping.txt

over what is in GeneOntology.py

PR #563, opened 2018-03-06
switched to  python built-in csv parser as ontobio gafparser no longe…

putmantime:
…r accepsts any none 'C,F or P' in aspect field


kshefchek:
For ECO codes, we may consider using the GO mapping file here: http://purl.obolibrary.org/obo/eco/gaf-eco-mapping.txt

See https://github.com/monarch-initiative/dipper/issues/437, and https://github.com/monarch-initiative/dipper/blob/077ac9672c3cbd9af19a620fa2644725899a115f/dipper/sources/Source.py#L480

putmantime:
Sounds good on the eco codes.  I could convert that to yaml and add it to dipper/translationtable

cmungall:
@dougli1sqrd I thought we were going to fix this so it's configurable

it's trivial to get the core info from a csv parser but you really don't want to do annotation extensions yourself

let's try extra hard to reuse code between monarch and go

putmantime:
Ok, @cmungall.  This was a quick and easy fix to get back RGD data.  But I can close this PR if updating ontobio to accept non go aspect values is in the pipeline

putmantime:
Great! Thanks @cmungall

PR #566, opened 2018-03-20
OMIM url updates, and named graph wip from lastyear

TomConlin:
The named graph wip stuff should stay out of the way as it is new and unused 
(famous last words)
The OMIM stuff should replace existing code, sooner is better.


kshefchek:
preliminary +1, @TomConlin can you fix the broken sparql tests?

TomConlin:
wonderful,
looks like we have improved the ingest to the point that all the test which should never have passed finally break.

Issue #564, opened 2018-03-14
Revise all prefixes on next release


cmungall:
I see we still have this:

```
"HGNC": "http://www.genenames.org/cgi-bin/gene_symbol_report?hgnc_id=HGNC:"
```

This is a terrible semweb PURL for multiple reasons. We should revise all PURLs and ensure we are at least synced with GO, @bgood @balhoff @kltm @dougli1sqrd 

See doc PURLs for bioentities: https://docs.google.com/document/d/1TvOERSc8rO8yXGFbnLvJjRHy6kApkvG3Gv98sLj43TE/edit

cmungall:
OK, it looks like GO-CAMs and Monarch are converging, we only have two clashes in prefixcommons: https://github.com/prefixcommons/biocontext/blob/master/reports/clashes-minerva-monarch.txt

PR #575, opened 2018-04-03
WIP: fix mychem, test refactor

kshefchek:
- [ ] fix mychem parser (removes limit)
- [ ] add tests to mychem
- [ ] refactor tests

The test refactor will remove the sparql tests in favor of testing exact sets of triples.  The new testing approach will be to:
1. Test that the graph object is empty
2. Run graph through function
3. Evaluate the set of expected triples vs the actual triples created by the function

This has the advantage over sparql in catching unexpected additional triples, where as the sparql queries would pass.  We should still use sparql for more integration style testing (see test in test_impc, test_random_data_set).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,5 +37,8 @@ RGD-test:
 SGD-test:
 	$(TEST) tests/test_sgd.py
 
+CTD-test:
+	$(TEST) tests/test_ctd.py
+
 trans-test:
 	$(TEST) tests/test_trtable.py
--- a/dipper/graph/RDFGraph.py
+++ b/dipper/graph/RDFGraph.py
@@ -48,7 +48,7 @@ class RDFGraph(ConjunctiveGraph, DipperGraph):
                     (self._getNode(subject_id), self._getNode(predicate_id),
                      Literal(obj)))
             else:
-                logger.warn(
+                logger.warning(
                     "None as literal object for subj: %s and pred: %s",
                     subject_id, predicate_id)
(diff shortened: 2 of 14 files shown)


TomConlin:
Looks good

putmantime:
+1

kshefchek:
@putmantime @mbrush are these the expected triples:
https://github.com/monarch-initiative/dipper/pull/575/files#diff-e36b4f838f2fc794fd34b42ed9ed2346R38
given the input:
https://github.com/monarch-initiative/dipper/pull/575/files#diff-f70e7770f0e3d56e1493d2635ea8579b

kshefchek:
Did I misinterpret https://github.com/monarch-initiative/dipper/issues/564? What should the iri be?

kshefchek:
we replaced the cgi-bin iri 3 years ago: https://github.com/monarch-initiative/dipper/commit/df76f391db8a72b4cec0354b2e6c8050344ab098 https://github.com/monarch-initiative/dipper/commit/5d62eaf7835ee4abd2f14984ac43845b40c284db

cmungall:
N2T wouldn't be an improvement. At least identifiers.org is used in a fair number of triplestores so one less join or rewrite is required in federated queries. The choice is between a firm ontological commitment in a new URI space or a non-ontological URI, see https://docs.google.com/document/d/1TvOERSc8rO8yXGFbnLvJjRHy6kApkvG3Gv98sLj43TE/edit

PR #590, opened 2018-06-07
Send request headers when checking response headers

kshefchek:
When checking response headers, make sure we are setting any request headers specified by the parser.  Fixes access issues with OMIM.


TomConlin:
+1   
perhaps make default for all requeats

PR #591, opened 2018-06-12
Update curie_map.yaml

vdancik:
Add CURIEs for Human Metabolite Database (http://www.hmdb.ca)

### DIFF
--- a/dipper/curie_map.yaml
+++ b/dipper/curie_map.yaml
@@ -151,6 +151,7 @@
 # pathways
 'KEGG-path': 'http://www.kegg.jp/dbget-bin/www_bget?path:'
 'REACT': 'http://www.reactome.org/PathwayBrowser/#/'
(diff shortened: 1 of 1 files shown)


TomConlin:
Looks fine but what are these to be used for?  

     'SMPDB': 'http://smpdb.ca/view/'
     'HMDB': 'http://www.hmdb.ca/metabolites/'


Are you creating an ingest?

vdancik:
We use the CURIE list for NCATS translator project (https://ncats.nih.gov/translator) and I am integrating metabolomics (HMDB) and small-molecule pathway (SMPDB) databases. So I am proposing to add those for completeness.

kshefchek:
While I'm not opposed, I'm not certain this file should represent the gold standard.  Would it make more sense to use prefix commons for this task?

jmcmurry:
Agree that for Translator purposes, these would be better to go in the Biolink curie map.

Issue #583, opened 2018-05-16
Parse sex-specific phenotypes in dipper


cmungall:
@pnrobinson is implementing a parser for MGI_Pheno_Sex.rpt from http://www.informatics.jax.org/downloads/reports/index.html

URL: http://www.informatics.jax.org/downloads/reports/MGI_Pheno_Sex.rpt

Should we re-use this? File seems fairly trivial to perhaps two parsers and coordinate on modeling?

Shall we have dipper follow the biolink model?

We already have a qualifier for sex:

https://biolink.github.io/biolink-model/docs/GenotypeToPhenotypicFeatureAssociation.html
https://biolink.github.io/biolink-model/docs/sex_qualifier.html

@mbrush can you quickly review? I assume me map M/F to http://purl.obolibrary.org/obo/PATO_0001894 subclasses?

cmungall:
Example row:

```
         Genotype ID: MGI:2167486
                 Sex: M
               MP ID: MP:0001147
             MP Term: small testis
 Allelic Composition: Kit<W-42J>/Kit<+>
   Background Strain: C57BL/6J-Kit<W-42J>/J
Sex-specific Normal Y/N: N
```

We have this already without an explicit sex qualifier (tho this can obv be inferred for this phenotype)
https://monarchinitiative.org/genotype/MGI:2167486#phenotypes

TomConlin:
other qualifiers I have come across recently include  

 - Sex unspecified
 - Sex ambiguous
 - Mixed sex

pnrobinson:
Our goal is to have a parser that will ingest the main MGI file, the sex-specific file, and the IMPC data to create an overall view of phenotype and sex specificity. So we would like to create one ingest tool in phenol that would just work with the main MGI data (this is the main use case) and have a second option that would allow use to combine this with the additional files.
Here is an issue (that needs to be tofu'd out)
https://github.com/monarch-initiative/phenol/issues/78

TomConlin:
related: https://github.com/monarch-initiative/phenol/issues/94

kshefchek:
Added with #595, #602

PR #595, opened 2018-06-20
Add sex specific G2P qualifier from MGI

kshefchek:
Updates the mgi db evidence query to include association metadata such as sex specificity - I wouldn't consider this evidence, but this is their representation.  For now I am hanging this off the association node.  See the unit test for an example input/output.

We are now getting back some other fields that may be useful, but are not added to the ingest in this PR:

Data Interpretation Center (eg IMPC)
external ref (eg UniProtKB:Q9JHI2-3 for Proteoform/Marker assoc)
Phenotyping Center (eg WTSI)
Resource Name (eg MGP)

Related ticket: https://github.com/monarch-initiative/dipper/issues/583

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,11 @@ TEST = python3 -m unittest
 ### Tests
 ###
 
-test: UDP-test IMPC-fetch IMPC-test GWAS-test reactome-test RGD-test \
+test: MGI-test UDP-test IMPC-fetch IMPC-test GWAS-test reactome-test RGD-test \
       string-test trans-test CTD-test mychem-test
+
+MGI-test:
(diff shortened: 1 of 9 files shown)


kshefchek:
I've also added in normal phenotypes, as these were being filtered out.  I may be missing something, but I couldn't find a reason to leave these out.  The ontology (MP, UPHENO) differentiates these as normal as opposed to abnormal.  Perhaps this had not been fully integrated in UPHENO when the original ingest was written?

kshefchek:
Test run and output from this branch:
http://ci.monarchinitiative.org/job/build-mgi-ttl/64/
https://data.monarchinitiative.org/ttl/mgi.ttl

TomConlin:
+1

PR #597, opened 2018-06-20
WIP: Align Impc sex specificity with MGI

kshefchek:
Aligns IMPC with MGI model for qualifying sex when applicable.  Technically redundant with the geno model which captures sex specific genotypes.


TomConlin:
+1

PR #598, opened 2018-06-25
omia test fix ; curie prefix defs

TomConlin:
- fixing omia tests by finding an example that exists
  - don't know where/why the last one disappeared
- capturing work finding definitions for curie prefixes


dipper/sources/Ensembl.py
```diff
@@ -230,15 +235,14 @@ def _build_biomart_gene_query(self, taxid):
...
 
-        object_attributes = {
-            "name": ensembl_taxon_to_db_map[taxid],
-            "interface": "default"
-        }
-        d = etree.SubElement(q, "Dataset", object_attributes)
-
         # TODO unused?
-        for i in cols_to_fetch:
-            a = etree.SubElement(d, "Attribute", {"name": i})
```
kshefchek: SubElement() appends data directly to an xml element, then returns a copy of the new element.  So even though it appears these are unused, removing this breaks the script.
https://docs.python.org/3/library/xml.etree.elementtree.html#xml.etree.ElementTree.SubElement

I'll change:

    a = etree.SubElement(d, "Attribute", {"name": i})

to 

    etree.SubElement(d, "Attribute", {"name": i})

TomConlin: thanks the fewer false negatives for static checking the better

PR #599, opened 2018-06-28
Automatically generate apidocs, consolidate docs in RTD

### LINKED ISSUE #490: Fix read the docs generation
Right now sphinx documentation is manually generated and tends to fall behind.  We should explore options for automating this process, for example: https://github.com/rtfd/readthedocs.org/issues/1139

kshefchek:
Autogenerates API docs using the method described here: https://github.com/rtfd/readthedocs.org/issues/1139

Also consolidates documentation that is currently spread out between two READMEs, notebooks, and different repos, as reStructuredText files.  Right now there’s some duplication but we should explore driving the docs from one format.  I briefly explored https://github.com/miyakogi/m2r and recommonmark.

Fixes #490


kshefchek:
Live version here: http://dipper.readthedocs.io/en/autodoc/

Issue #593, opened 2018-06-18
Use ClinVar curated gene to disease file for G2P associations


kshefchek:
ClinVar provides a curated gene to disease file here:
ftp://ftp.ncbi.nlm.nih.gov/pub/clinvar/gene_condition_source_id

This will help resolve cases where a variant affects more than one gene, where one or more genes should not be propagated to the disease:
https://www.ncbi.nlm.nih.gov/clinvar/docs/faq/#manygenes

Example issues with the current implementation
MIR3606 incorrectly associated with EDS type 4: https://monarchinitiative.org/disease/MONDO:0007524#genes
https://github.com/monarch-initiative/monarch-app/issues/1591

We could use this file to infer that a variant should be associated with a gene via the has_affected_feature relation versus has_reference_part, although perhaps this is a modelling hack (@mbrush can you advise?).  Alternatively, we could use this file to link genes and diseases directly, and adjust our gene to disease inferences through variants.

Ideally we would have a source of variant to gene associations so we would not need to make this inference.  We should explore if such a source is available.

pnrobinson:
Interesting! 
Does anyone know what the relation is to OMIM-based resources? It looks like the ClinVar resource has more, but is it a strict superset of mim2gene_medgen?

kshefchek:
The last time I looked into it there were conflicts between OMIM and ClinVar, see https://github.com/monarch-initiative/monarch-cypher-queries/issues/23#issuecomment-345119000

kshefchek:
Added with #602

kshefchek:
This isn't quite fixed - see https://monarchinitiative.org/disease/MONDO:0007947#gene-other

kshefchek:
fixed with https://github.com/monarch-initiative/dipper/pull/729

Issue #560, opened 2018-02-01
Simplify modeling patterns and eliminate blank nodes that do no work for us


cmungall:
Dipper should not introduce blank nodes unless either they are present in the structure of the source or if it does work for us in reasoning. The structure of the graph should reflect the current solr 'triples'.

Even in cases where these are introduced, we should place a shortcut relation.

This would need to be coordinated with https://github.com/monarch-initiative/monarch-cypher-queries/tree/master/src/main/cypher/golr-loader (the queries would become simpler as will be an isomorphic triple to triple mapping)

TomConlin:
+1,000,000

kshefchek:
> Dipper should not introduce blank nodes unless either they are present in the structure of the source or if it does work for us in reasoning

I think we need to clearly define what this means.  I'm guessing our most common use case for bnodes are instances where a source provides a label but does not expose a persistent ID.

jmcmurry:
I believe it is also applied when the ID is persistent but not resolvable;
Tom can you comment?

TomConlin:
I intentionally did not mention our _not-even_-bnode identifier constructs created as IRI for under identified external objects to avoid expanding the scope of this issue. (but they are the first thing that come to mind)

one thought may be to establish a curie  `LCL:` (shades of the fasta defline standard) which 
denote what follows is a __LOCAL__ identifier with disclaimers to what that means

jmcmurry:
What happened to our plan to create surrogate IDs in this stable-but-not-resolvable scenario?

TomConlin:
I do not know, perhaps it was another thought NW was not able to communicate before she left?

cmungall:
Sorry I wasn't precise in my original wording. Let's keep this ticket about modeling and not identifiers. I'm talking about the pattern of introducing a blank or otherwise surrogate node for an entity that is only implicit in the source.

The broader context is simplifying things for bioinformaticians. If someone is in front of their computer using either the neo4j interface or the scigraph swagger, and they have a gene `G`, and they want to find a phenotype `P` that is somehow associated, then their query should be `SELECT * WHERE {G ?R ?P}` (or whatever the cypher equivalent), not

```cyphr
      MATCH (locus:gene)<-[:GENO:0000418!]-(feature)
        WITH feature, COUNT(DISTINCT(locus)) as gene_count
        WHERE gene_count = 1
        AND NOT feature:snv
[...]

Issue #547, opened 2017-11-17
Pull in drug response phenotypes from ClinVar


kshefchek:
For example:
https://www.ncbi.nlm.nih.gov/clinvar/variation/722/ - NAT2 variant and OMIM:243400 (Slow Acetylation)

We could pull these in if there is an associated OMIM or medgen identifier.

TomConlin:
[ClinVarTestSet.xml.gz](https://github.com/monarch-initiative/dipper/files/1483579/ClinVarTestSet.xml.gz)

Issue #459, opened 2017-05-18
Clinvar introduced new elements for 'CompoundHeterozygote'


TomConlin:
ftp://ftp.ncbi.nlm.nih.gov/pub/clinvar/release_notes/20170504_data_release_notes.pdf

```

The May release uses a new element, GenotypeSet, to indicate that an interpretation was made about
two alleles that are on different chromosomes.This case was formerly reported as:
<MeasureSet Type="CompoundHeterozygote" ID="424841">
and is now reported as:
<GenotypeSet Type="CompoundHeterozygote" ID="424841">
    <MeasureSet Type="Variant" ID="375249">
        <Measure Type="single nucleotide variant" ID="361947">...</Measure>
    </MeasureSet>
    <MeasureSet Type="Variant" ID="375250">
        <Measure Type="single nucleotide variant" ID="361945">...</Measure>
    </MeasureSet>
</GenotypeSet>
Currently all GenotypeSets are of Type "CompoundHeterozygote", indicating a simple allele on each
chromosome. In the future we will also include GenotypeSets of Type "Diplotype", indicating a
haplotype on one or both chromosomes.
```

kshefchek:
Per #420, we should model complex hets in this way:

    ClinVarVariant:424841 a GENO:0000030 ;
        GENO:0000608 GENO:0000402 ;
        GENO:0000382 dbSNP:80159221 ;
        GENO:0000382 dbSNP:778624615 ;
        GENO:0000845 OMIM:208900 .


Where
GENO:0000030 ! variant single locus complement
GENO:0000608 ! has_zygosity
GENO:0000402 ! complex heterozygous
GENO:0000382 ! has_variant_part
GENO:0000845 ! has_uncertain_significance_for_condition

pnrobinson:
What does phrase "complex heterozygous" mean? Is this the same as "compound heterozygous" or intended to be a generalization?

kshefchek:
From GENO: a heterozygous quality inhering in a locus complement comprised of two different varaint loci and no wild type locus. (e.g.fgf8a&lt;ti282a&gt;/fgf8a&lt;x15&gt;)

pnrobinson:
In medical genetics, we would use "compound heterozygous" for this, and I have never heard complex heterozygous -- unless there is a reason for the label (e.g., used by model organisms?), it probably should be changed!

kshefchek:
[...]

PR #602, opened 2018-07-06
Improve ingests and modelling of human g2p data

kshefchek:
### Background
There are issues with our representation of associations between genes, variants, and diseases.
- Over use of blank nodes
- Incorrect gene to disease inferences, eg https://github.com/monarch-initiative/monarch-app/issues/1591, 

#### New Sources
EBI Gene2Phenotype: https://www.ebi.ac.uk/gene2phenotype

#### Updated Sources 
OMIM, ClinVar, Orphanet, CTD

#### Modelling considerations:

No single source offers a comprehensive structured source of variant to disease associations (much is still in free text).

Variant attributes (Add if not currently ingested):
Type: eg SNP, indel (ClinVar)
Origin: eg Germline, somatic (Orphanet)
Molecular consequence: eg nonsense, missense (ClinVar)
Functional consequence: eg loss of function (Orphanet, EBI)
Location:
  - genomic (ClinVar)
  - cytogenic (OMIM, ClinVar, Orphanet)

Genotype attributes (No changes):
Zygosity (hemi, homo, hetero, compound het) (No sources, but in free text descriptions)

Disease attributes (No changes):
Inheritance (HPOa, OMIM)

Qualifiers on a variant-disease association:
Allelic requirement (eg monoallelic, biallelic) could also argue an attribute of a genotype
(ie an allelic genotype with two loss of function variants of gene X) (EBI)

Associations between variants and their affected genes should be captured in the modelling - see https://github.com/monarch-initiative/dipper/issues/593

We want to remove blank nodes in cases where they do not serve a function from a modelling perspective, for example, most 
variant blank nodes in OMIM and CTD.  Cases where we may consider blank nodes:

- where a disease and genomic consequence are conflated, for example: https://omim.org/entry/400045
- where we have attributes of a variant, but no identifier (EBI, Orphanet), we could also move these to the association node, but then we're technically trading one blank node for another.

Source updates:
Orphanet - new relations: Biomarker tested in, Candidate gene tested in
Replace DisorderGeneAssociationType IDs with label translation table mappings

OMIM - Remove most variant bnodes (see above for exceptions)

CTD - Remove all variant bnodes

ClinVar - WIP

#### Relevant issues:
https://github.com/monarch-initiative/dipper/issues/593
https://github.com/monarch-initiative/dipper/issues/547
https://github.com/monarch-initiative/dipper/issues/459
https://github.com/monarch-initiative/dipper/issues/560


dipper/sources/EBIGene2Phen.py
```diff
@@ -0,0 +1,315 @@
...
+        # Load mondo map file
+        self.mondo_map = {}
+        map_url = self.map_files['mondo_map']
+        with closing(requests.get(map_url, stream=True)) as map_fh:
+            for line in map_fh.iter_lines():
+                row = line.decode('utf-8').split('\t')
+                self.mondo_map[row[0]] = row[1]
+
+        self.global_terms = self.open_and_parse_yaml('../../translationtable/global_terms.yaml')
+        self.translation_table = self.open_and_parse_yaml('../../translationtable/ebi.yaml')
```
TomConlin: Going to ask the filename be the same base that is used elsewhere   
which is the ingest file, name lowercased,  sans .py 

`ebigene2phen.yaml` here

dipper/sources/MPD.py
```diff
@@ -453,6 +454,12 @@ def _add_g2p_assoc(self, g, strain_id, sex, assay_id, phenotypes, comment):
                 assoc.add_association_to_graph()
                 assoc_id = assoc.get_association_id()
                 model.addComment(assoc_id, comment)
+                if sex == 'm':
+                    model._addSexSpecificity(assoc_id,
+                                             self.global_terms['male'])
+                elif sex == 'f':
+                    model._addSexSpecificity(assoc_id,
+                                             self.global_terms['female'])
```
TomConlin: this could be set up with local  having `'m': 'male'`   &  `'f' : 'female' ` 
then the stanza becomes  the line:  
 
     model._addSexSpecificity(assoc_id, self.resolve(sex))

tests/resources/global_terms.yaml
```diff
@@ -0,0 +1,11 @@
+# Creating a separate global terms so in case something
+# is updated the test does not break
```
TomConlin: not like.

how about `test_terms.yaml`

tests/test_orphanet.py
```diff
@@ -25,5 +25,140 @@ def tearDown(self):
...
+
+<https://monarchinitiative.org/.well-known/genid/b56f798350412a34> a GENO:0000002 ;
+    rdfs:label "germline variant of KS1" ;
+    GENO:0000418 Orphanet:268061 ;
+    RO:0003303 Orphanet:938475 ;
+    :MONARCH_anonymous true ;
+    :has_cell_origin GENO:0000900 .
+
+Orphanet:938475 a owl:Class ;
+    rdfs:label "too much unit testing disorder" .
```
TomConlin: +1

translationtable/ebi.yaml
```diff
@@ -0,0 +1,123 @@
+# See https://www.clinicalgenome.org/site/assets/files/2757/fitzpatrick_ddg2p.pdf
```
TomConlin: Want name to local tt name to be a common function of ingest name it is local to

translationtable/orphanet.yaml
```diff
@@ -0,0 +1,53 @@
+# Orphanet stores variant-disease associations
```
TomConlin: maps of maps.  had not been expecting that.  
there is a function `Source.resolve()`  that will not know what to do with this.
I see it solves a scoping problem, that will be good to keep solved,  
 but I will have to think about the down stream consequences.

and precomposing the keys: 
`"Biomarker tested in gene phenotype": "is_marker_for"`
`"Disease-causing germline mutation(s) (gain of function) in gene phenotype": "causes_condition"`

is cumbersome in a different way

TomConlin:
I am happy to hear arguments in favour of translation tables including maps-of-maps but do not find sufficient cause myself.  the space they save and programmer convenience they offer are not out weighing  clarity, inverting, validation, survey/reporting & maintenance.

kshefchek:
Makes sense, how would I go about modelling the orphanet translation table as a map?  I could make the values a list or tuple but then it's less clear for whoever maintains this.  And I'm not sure that resolves the issue, is the requirement that the values are scalar?

kshefchek:
something like this?

    "Disease-causing germline mutation(s) (gain of function) in-functional_consequence": "gain_of_function_mutation"

TomConlin:
looks good to me

kshefchek:
@TomConlin I think I've resolved all of your suggestions, with the exception of https://github.com/monarch-initiative/dipper/pull/602#pullrequestreview-141249305.  I still need to add tests for ebi g2p and clinvar but I'm tempted to get this merged so we're developing on the same branch.

kshefchek:
Github is having issues with serving files, so some ontology files are returning 502s, see https://status.github.com/messages, will wait till it resolves.

TomConlin:
+1

PR #605, opened 2018-07-20
small fixes

TomConlin:
- GO had a novel curie
 - cause Genotype class to fail if a 'None' get that far
 - coriell formatting
 - maven dir does not belong here


kshefchek:
Let's rehome before deleting, I think this is still used in the scigraph load process to configure curies.  I can check.

TomConlin:
if scigraph is using it we should figure out how to have the maven bits live there (or just fetch the curie_map)

kshefchek:
It's a monarch specific configuration so it could live in one of our config repos -  scigraph-docker-monarch, or configs.  It is just yaml file construction, we may consider rewriting in some scripting language.  Having everything in that pipeline in maven is nice, but something non-xml based could be more transparent.

TomConlin:
I was looking in SciGraph-docker-monarch-[data|ontology] 
still looking for where the dipper maven dir is actually referenced

What do you mean by yaml file construction? 
curie_map.yaml is obviously already a yaml file

kshefchek:
This is for constructing the scigraph configuration yaml.  It looks like it's hidden away in the jenkins job:

# Get and install the queries and curies
git clone https://github.com/monarch-initiative/dipper.git /tmp/dipper
git clone https://github.com/monarch-initiative/monarch-cypher-queries.git /tmp/monarch-cypher-queries
git clone https://github.com/SciGraph/SciGraph.git /tmp/SciGraph

cd /tmp/SciGraph && mvn clean install -DskipTests
cd /tmp/dipper/maven && mvn clean install
cd /tmp/monarch-cypher-queries && mvn clean install

+1 for cleaning this up

TomConlin:
How would you feel about putting it in `monarch-cypher-queries.git` or is that more of a step sideways

kshefchek:
Yes feels a little sidestepy,  maybe we should use a yaml parser, but this could easily be done in sed or awk:

In [monarchConfiguration.yaml.tmpl](https://github.com/monarch-initiative/SciGraph-docker-monarch-data/blob/master/src/main/resources/monarchConfiguration.yaml.tmpl):

      curies:
    INJECT_CURIES

get curie map and add an indent:

    wget https://raw.githubusercontent.com/monarch-initiative/dipper/master/dipper/curie_map.yaml
    sed -i 's/^/    /' curie_map.yaml

Replace with awk:

    awk -v replacement="$(<curie_map.yaml)" '
    {gsub("INJECT_CURIES", replacement); print}
    ' monarchConfiguration.yaml.tmpl

TomConlin:
you know that I will approve of this message. 

And it reminds me 
I did see some promise with the git ignore approach 
for getting jenkins under source control 
and now that I am wantonly breaking things 
I should revisit that

PR #607, opened 2018-07-27
AnimalQTL fix and translation tables in Source.py

TomConlin:
- Animal QTL
  - break out the curie/url for each species
  - dataset changes per last PR

- Source.py  
  - Have the ingest super class Source house the global and local translation tables 
  - include a function to resolve keys

### DIFF
--- a/dipper/sources/AnimalQTLdb.py
+++ b/dipper/sources/AnimalQTLdb.py
@@ -94,15 +94,13 @@ class AnimalQTLdb(Source):
 
     # QTL ids
     test_ids = {
-        28483, 29016, 29018, 8945, 29385, 12532, 31023, 14234, 17138, 1795,
-        1798, 32133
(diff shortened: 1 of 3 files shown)


dipper/sources/AnimalQTLdb.py
```diff
@@ -337,8 +337,8 @@ def _process_QTLs_genetic_location(
                     reference = Reference(g, pub_id)
                 elif pubmed_id != '':
                     pub_id = 'PMID:'+pubmed_id.strip()
-                    reference = Reference(g,
-                        pub_id, Reference.ref_types['journal_article'])
+                    reference = Reference(
+                        g, pub_id, Reference.ref_types['journal_article'])
```
kshefchek: In the spirit of the o->common_name refactor above, change g to graph.

kshefchek:
+1

dipper/sources/AnimalQTLdb.py
TomConlin: I can't even guess what 'k' is ... maybe 'kind'?

PR #610, opened 2018-08-24
Translation table refactor (in process, pulling to run ingest on Jenkins)

TomConlin:
big changes, 
~5k loc edited in over 100 files.  
net loss of ~1k locs
not exactly realistic to expect a comprehensive review right away and my main goal is 
to be running the next set of testing 'not on my laptop'.

main points are:
 - All ontology terms emitted from every ingest,  are drawn from a single yaml file to ensure transparency and consistently downstream. 

- Each ingest has its own local translation table to map the sources _whatever_ to the target terms in the global mapping
- Each local translation table may also a double as a way to transform literals 

- The global translation table `globaltt` is automatically available in all files (as is the global curie_map)

- The local translation table `localtt` is automatically available in every source ingest
  - a helper function `resolve(word')` is also available  in source ingests which uses both the local and global translation tables together for when the input is messy.
 
- inverse translations are is also available because using them can help
  - code readability (see common terms instead of ingest specific wordage)
  - write meaningful logging messages e.g; what you have to do the fix this error


dipper/sources/EBIGene2Phen.py
```diff
@@ -138,7 +137,7 @@ def parse(self, limit: Optional[int]=None):
 
         return
 
-    def _add_gene_disease(self, row:List):
+    def _add_gene_disease(self, row):  # ::List  getting syntax error here
```
kshefchek: In python 3.5?

TomConlin:
``` 

cat Pipfile
[[source]]
name = "pypi"
url = "https://pypi.org/simple"
verify_ssl = true

[packages]
rdflib = "*"
isodate = "*"
roman = "*"
requests = "*"
PyYAML = "*"
psycopg2 = "*"
pandas = "*"
sparqlwrapper = "*"
pysftp = "*"
ontobio = "*"

[requires]
python_version = "3.5"

[dev-packages]
(dipper-BclvAZdW) tomc@cypher ~/Projects/Monarch/dipper $ python --version
Python 3.5.2

```
yep

TomConlin:
I will avoid building GWAS and Orphanet till last in case anyone wants to take a run at fixing the testd

cmungall:
Refactor looks great!!

Did you consider using vocab objects for increased strong typing for some of the lookups, e.g. https://github.com/biolink/ontobio/tree/master/ontobio/vocabulary

TomConlin:
no want them as plain and simple as possible and use ontobio offline to help marshal the ontology teams.  helps keep them usable in unforeseen ways

TomConlin:
@cmungall  just to be a little more clear on my  "no" above, 
we want the same thing just have different approaches
you took the OO path of encoding your filter constraint in objects 
which will require maintenance as ontologies evolve .

I want to generate partitions of the global tt mapping 
based on what their ontology says they are now.
If in the code we know a term should be of a particular ilk
we can query the particular generated map partition and then 
decide what to do on a miss, revert the the global file or fail hard etc.

cmungall:
Ok!

PR #671, opened 2018-10-15
HPOA and related

### LINKED ISSUE #469: Coordinate on orphanet prefix
We use 'Orphanet' in dipper, which is consistent with MonDO, which is consistent with ORDO produced by Orphanet + the EBI

It looks like the official portal uses 'ORPHA'
http://www.orpha.net/consor/cgi-bin/OC_Exp.php?Expert=597

(I think this is new?)

If we change to ORPHA it will take careful coordination between dipper and mondo and releases.

Alternatively we could leave as-is and document the difference, e.g syn if prefixcommons/db-xrefs.yaml

cc @drseb @jmcmurry

TomConlin:
Goal was to use the new HPO files. 
Along the way the common-disease files raised enough concerns they are not currently included and a new push for a common disease model & curation will begin RSN.

- dependency on python git was dropped
- Disposition association was merged into the D2P association  (nothing distinct but relation)
- lots of convention
  - continue rooting out single letter variable names
  - change variable names which are reserved words in language
  - staunch the use of java idioms in favor of python idioms (not claiming either better)
- reduce the use of the regular expression engine to test for exact equality
- adopt Orphanet's preferred curie prefix `ORPHA:` closes #469

There is a problem with the abstract method `serialize` in Graph having a different signature than the same named method in the rdflib.  where Source and GraphUtils
each expect a different implementation.  
The current workaround is to neglect to implement the serialize method in RDFGraph.
This should be addressed before it comes up in practice.


kshefchek:
+1, we should circle back on common disease and include it as a separate ingest

Issue #679, opened 2018-10-26
dipper/graph/RDFGraph.serialize()


TomConlin:
`dipper/graph/Graph` is an abstract class and has an abstract method `serialize()`

`dipper/graph/RDFGraph`  has multiple inheritance from  `dipper/graph/Graph` and `rdflib/ConjunctiveGraph`  
 which also implements a method `serialize()` with a different signature.
which keeps `dipper/graph/RDFGraph` technically abstract as it is not also implementing  
`dipper/graph/Graph.serialize()`

resolve such that instances of  `dipper/graph/RDFGraph` may used   
without `pylint3` issuing an error.

TomConlin:
will try that.

If so, we must be accustomed to eating them bugs by now.
It causes static analysis tools to issue their sternest protests.

    sudo apt-get install pylint3

PR #680, opened 2018-10-27
reconcile multiple inheritance 'serialize()' ambiguity 

TomConlin:
trading pylint3 (version 1.5.2)  Error reports for   
instantiating a partially abstract class (due to one parent but not the other)
for pylint3 warning for inconsistent method signatures 

wanting this to go in before a larger pull of refactor cleanup noise


TomConlin:
closes #679

PR #691, opened 2018-12-20
neither curie nor IRI  as source are literals

### LINKED ISSUE #681: IRIs in hpoa.ttl are mistakenly string values
I downloaded `https://data.monarchinitiative.org/ttl/hpoa.ttl`. It can't load properly as OWL, because (like other dipper files) it states that `dc:source` is an object property, however there are many `dc:source` values in this file which are strings.

I assume all these strings are supposed to actually be IRIs:

```turtle
:MONARCH_b00012161a6f63bad72e a OBAN:association ;
    OBO:RO_0002558 OBO:ECO_0000501 ;
    dc:source "http://omim.org/entry/148050" ;
    OBAN:association_has_object OBO:HP_0008438 ;
    OBAN:association_has_predicate OBO:RO_0002200 ;
    OBAN:association_has_subject OBO:OMIM_148050 .
```

The `dc:source` property has IRI values in other dipper files.

cc @shalsh23

TomConlin:
resolves #681

### DIFF
--- a/dipper/models/Model.py
+++ b/dipper/models/Model.py
@@ -66,8 +66,7 @@ class Model():
                 class_id, self.globaltt['label'], label, object_is_literal=True)
 
         if class_type is not None:
-            self.graph.addTriple(
-                class_id, self.globaltt['subclass_of'], class_type)
+            self.graph.addTriple(class_id, self.globaltt['subclass_of'], class_type)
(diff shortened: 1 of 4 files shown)


TomConlin:
note: travis checks are breaking on basic http requests

PR #694, opened 2019-01-05
Biogrid & Reactome test passing

TomConlin:
- global effects in how curies are recognized (more lenient)
- global change in test id availability in unit tests (exist now)
- file sizes are now zero when they were None or empty
- Biogrid included in Makefile unit test suite


TomConlin:
(merge test fail is network time out)

PR #695, opened 2019-01-07
MMRRC fix ++

TomConlin:
fix bad data. MGI curie missing the 'I'  
however since I was in to find & fix that a bunch of other refactoring happened as well.

mmrrc
 - table column headers conform to the new practice now
 - logging reduced, just a summary at the end  when a symbol is missing an identifier  
 - remove some of the scrubbing for issues not happening now/anymore  
 - replace hard coded identifiers with translation table lookups
 -  reduce using regexps  for plain string slices.

 Source:
 - eliminate the reuse/clobber of 'row' as both a list and a dict in a loop
 - comments on vagueness of a very specific parser  (MySQL_XML dump)
    - needs revisiting

scripts/mysql2sqlite
 - script from GitHub to transform MySQL dumps into Sqlite syntax I use

### DIFF
--- a/dipper/sources/MMRRC.py
+++ b/dipper/sources/MMRRC.py
@@ -40,13 +40,32 @@ class MMRRC(Source):
     files = {
         'catalog': {
             'file': 'mmrrc_catalog_data.csv',
-            'url': 'https://www.mmrrc.org/about/mmrrc_catalog_data.csv'},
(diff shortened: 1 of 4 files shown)


TomConlin:
noting again that making impc[fetch/test] do not see a network time out from my machine

PR #697, opened 2019-01-11
HPOA Reference sanity check

TomConlin:
reject references as objects which are not well formed curies or URLs

@kshefchek  asking you to review to point out the use of `graph.curie_regexp`
available most everywhere in dipper.

If we ever start running into legitimate UTF8 curies it will need extending.

### DIFF
--- a/dipper/sources/HPOAnnotations.py
+++ b/dipper/sources/HPOAnnotations.py
@@ -163,6 +163,8 @@ class HPOAnnotations(Source):
         :return:
 
         """
+        src_key = 'hpoa'
+
         if self.test_mode:
             graph = self.testgraph
         else:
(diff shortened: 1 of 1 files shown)


kshefchek:
While you are working on hpoa can you filter out negated phenotype associations, eg

    if row[col.index('Qualifier')] == "NOT":
        continue

dipper/curie_map.yaml
```diff
@@ -147,7 +147,7 @@
 
 # variants / traits
 
-# 'AQTL-BT': https://www.animalgenome.org/cgi-bin/QTLdb/BT/search       # maybe?
+# 'AnimalQTLdb': 'previously'
```
kshefchek: Is this intentional?

TomConlin: Yes, 'AnimalQTLdb' existed as a curie prefix in the past.
it  was broken and did not go anywhere useful but   
I wanted to leave a breadcrumb in case anyone ever came looking for it.

PR #702, opened 2019-02-06
OMIM: adopt Phenotypic Series curie 

### LINKED ISSUE #701: OMIM phenotypic series IRIs incorrect
Currently: OMIM:PS1234, which expands to

https://www.omim.org/entry/PS309530

They should link to:
https://www.omim.org/phenotypicSeries/PS309530

Perhaps we should have a new entry in the curie map,
OMIMPS: https://www.omim.org/phenotypicSeries

TomConlin:
OMIM was built using the same curies for omim identifiers and Phenotypic Series identifiers
this PR corrects that.  closes #701

Movement towards codifying the expected incoming headers 
(as has happened across many ingests already)
is also captures in a pre functional state.
These column headers are used to detect/warn of incoming churn and  to 
mitigate code changes for simple column rearrangement/add/drop
(unless a dropped  column is explicitly required) 
and to preserve the source’s intent  for the data our code is acting on.

### DIFF
--- a/dipper/sources/GeneOntology.py
+++ b/dipper/sources/GeneOntology.py
@@ -17,7 +17,7 @@ from dipper.models.Genotype import Genotype
 from dipper.models.Reference import Reference
 from dipper.models.Model import Model
 from dipper.utils.GraphUtils import GraphUtils
-from dipper import config
+
 
 LOG = logging.getLogger(__name__)
 GOGA = 'http://geneontology.org/gene-associations'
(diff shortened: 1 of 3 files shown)


kshefchek:
Should the curie_map.yaml file be updated too (possibly forgot to git add) ?

TomConlin:
yes, it looks like I diffed them together but add separately (half), which is not my normal.
good catch

PR #712, opened 2019-02-26
Add jenkins pipeline for dipper

kshefchek:
Jenkins declarative pipeline file for dipper that fetches and parses sources and adds rdf files to data.monarch.  Also runs the monarch-merged owl job.


TomConlin:
We should go over the sed interventions.  some may not be needed any more
but adding this and evolving it is a great step in the right direction 
+1

PR #713, opened 2019-02-26
String db

TomConlin:
- String
   - Merge my changes into Kent's which helps necessary files to be found 
   - Note they are very sensitive to query rate. test frugally
   - Include Yeast
 - leftover ancillary scripts for ingest qc

### DIFF
--- a/dipper/models/Model.py
+++ b/dipper/models/Model.py
@@ -212,7 +212,7 @@ class Model():
 
     def addXref(self, class_id, xref_id, xref_as_literal=False):
         self.graph.addTriple(
-            class_id, self.globaltt['has_dbxref'], xref_id,
+            class_id, self.globaltt['database_cross_reference'], xref_id,
             object_is_literal=xref_as_literal)
         return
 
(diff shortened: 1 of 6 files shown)


kshefchek:
+1

>     * Note they are very sensitive to query rate. test frugally

Locally I had hacked in a sleep in between calls in Source.py, but ideally this would be some optional param.

TomConlin:
@kshefchek  what fate for https://data.monarchinitiative.org/dipper/cache/

TomConlin:
Something is different when running locally v.s. on jenkins causing the parse() function to attempt to process files the fetch() function have not retrieved.
I thought it odd Kent would push a script that did not run but mine is doing no better.

clearing my local 'raw/string/' directory and running works locally
the directory is populated them processed.
on jenkins parse attempts to process the empty directory

TomConlin:
mystery solved

the Jenkins script contained a directive to ` --parse_only`
most likely b/c failures due to StringDBs anti server overload policies

PR #722, opened 2019-04-10
increased granularity of mapping when ingesting Orphanet

justaddcoffee:
Changes in orphanet.yaml to increase granularity of mapping when ingesting Orphanet


kshefchek:
@justaddcoffee this looks good to me.  Previously we used a blank node to attach data such as cell origin and functional consequence, but since these are encoded in the relation they can be removed.  def add_gene_to_disease() and _add_variant_attributes() can be deleted, and at https://github.com/monarch-initiative/dipper/blob/4e1235/dipper/sources/Orphanet.py#L179 we would add:

    assoc = G2PAssoc(
            self.graph, self.name, gene_curie, disease_id, rel_id)
    assoc.add_evidence(eco_id)
    assoc.add_association_to_graph()

justaddcoffee:
Okay, makes sense. @kshefchek in that code snippet, where is rel_id coming from? Is that dg_label?

kshefchek:
You can retrieve any identifier for your term by using self.resolve("some label"), where the label refers to data in the source (Orphanet in this case) and it has an associated mapping in the yaml file.  Right now there is a two way mapping from source label -> ontology term -> ontology curie.  I'm guessing if you uncomment this it may work:

    rel_id = self.resolve(dg_label)

justaddcoffee:
Thanks @kshefchek! 

    `rel_id = self.resolve(dg_label)`

does work, but generates this warning:

    `INFO:root:Translated to 'is causal germline mutation in' but no global term_id for: 'Disease-causing germline mutation(s) in'`

and then this error when calling assoc.add_association_to_graph():

   `*** ValueError: ('Invalid Predicate for this association <%s> <%s> <%s>', 'HGNC:30497', 'is causal germline mutation in', 'ORPHA:166024')`

I'm probably not doing something correctly in the translation table (?)

kshefchek:
This is the first time we're using the new RO terms, so they'll need to be added here:
https://github.com/monarch-initiative/dipper/blob/master/translationtable/GLOBAL_TERMS.yaml

justaddcoffee:
Awesome, thanks! Is there an exhaustive list of new RO terms somewhere? I could just add the one I need, but assuming it'd be better to add all the new ones and not just the one I need here....

TomConlin:
Hi Justin . no adding those term is not what this is for. 
this is the topic of out discussion later today

justaddcoffee:
Okay @TomConlin, chat about this later then

justaddcoffee:
Okay, this PR is possibly ready now. Increases granularity of orphanet ingestion, passes make tests, and ./dipper-etl.py -s Orphanet runs apparently correctly. Also removed some superfluous code and did some trivial tidying up

test.py
```diff
@@ -0,0 +1,38 @@
+import pandas as pd
```
kshefchek: Was this file meant to be committed?

translationtable/orphanet.yaml
```diff
@@ -1,13 +1,18 @@
...
+# (RO:0004015)
+"Major susceptibility factor in": "is causal susceptibility factor for"
+# (RO:0004016)
+"Modifying germline mutation in": "is causal germline mutation partially giving rise to"
+
+# EBI seems to have removed "Modifying somatic mutation in" - strange, because "modifying germline mutation in" (above) still exists.
+# https://www.ebi.ac.uk/ols/search?q=Modifying+germline+mutation+in&groupField=iri&exact=on&start=0&ontology=ordo&type=property
+
+"Part of a fusion gene in": "contributes to" # I don't see any appriopriate RO term for this that is more granular
+"Role in the phenotype of": "contributes to" # this seems like the right granularity here
```
kshefchek: I think we should use "contributes to condition" instead of "contributes to"

test.py
justaddcoffee: No, sorry, that was a test script hanging out in that directory - that's what I get for doing a `git add *`

dipper/sources/Orphanet.py
```diff
@@ -190,160 +169,3 @@ def _process_diseasegene(self, limit):
...
-            eco_id):
-        """
-        Composes triples based on the DisorderGeneAssociationType element:
-        AND the suffixes:
-
-            - "gene phenotype"
-            - "function consequence"
-            - "cell origin"
-
-        xmlstarlet sel -t  -v "/JDBOR/DisorderList/Disorder/DisorderGeneAssociationList/
```
kshefchek: I think we should keep this command somewhere in the script, perhaps in a different docstring

translationtable/orphanet.yaml
justaddcoffee: agreed, that seems better, just pushed a commit with this change

dipper/sources/Orphanet.py
justaddcoffee: Ah okay, just added it below the first docstring for now, so at least it's not lost to the ages...

translationtable/GLOBAL_TERMS.yaml
```diff
@@ -485,7 +486,6 @@
 "ends_with": "RO:0002230",
 "has_input": "RO:0002233",
 "colocalizes with": "RO:0002325",
-"contributes to": "RO:0002326",
```
kshefchek: Let's double check if this affects any other sources, this was used occasionally before "contributes to condition" was added to RO.  Possibly used in GWAS, OMIM

justaddcoffee: Added "contributes to": "RO:0002326" back at least for now

Makefile
```diff
@@ -83,17 +83,18 @@ clean_prefix_equivalents:
 
 translationtable/generated/curiemap_prefix.txt: dipper/curie_map.yaml
 	@ cut -f1 -d ':' dipper/curie_map.yaml  | tr -d "'" | egrep -v "^$|^ *#" |\
-		grep .|sed 's|\(.*\)|"\1"|g'| LANG=en_EN sort -u > \
-			translationtable/generated/curiemap_prefix.txt
+		grep .|sed 's|\(.*\)|"\1"|g'|\
```
TomConlin: back slash is implied after pipe

Makefile
```diff
@@ -83,17 +83,18 @@ clean_prefix_equivalents:
...
-	@ LANG=en_EN join translationtable/generated/curiemap_prefix.txt  /tmp/local_inverse.tab  |\
-		awk '{v=$$1;$$1="";print $$0 ": " v}'| sort -u  > \
-			translationtable/generated/prefix_equivalents.yaml
+				translationtable/[a-z_-]*.yaml |\
+		LC_ALL=C sort -u > /tmp/local_inverse.tab
+
+translationtable/generated/prefix_equivalents.yaml: \
+		translationtable/generated/curiemap_prefix.txt /tmp/local_inverse.tab
+	@ LC_ALL=C join translationtable/generated/curiemap_prefix.txt \
+		/tmp/local_inverse.tab | awk '{v=$$1;$$1="";print $$0 ": " v}'|\
```
TomConlin: ditto

justaddcoffee:
> what are all my changes doing in your PR?
Some merge strangeness, will have a look tomorrow

justaddcoffee:
> what are all my changes doing in your PR?

Rebased to remove your commits from my PR - should be correct now (although the comments in this thread are now a bit confusing).

kshefchek:
I'm curious if these still exist in Orphanet:
Disease-causing germline mutation(s) (loss of function) in
Disease-causing germline mutation(s) (gain of function) in

TomConlin:
@justaddcoffee are you seeing any evidence that the code changes are having any effect on the output?  perhaps I am tired and being sloppy but I am not seeing any increase in granularity in the results

justaddcoffee:
> I'm curious if these still exist in Orphanet:
> Disease-causing germline mutation(s) (loss of function) in
> Disease-causing germline mutation(s) (gain of function) in

@kshefchek Do you mean we should get rid of these from the local translation table if they aren't being used?

justaddcoffee:
> @justaddcoffee are you seeing any evidence that the code changes are having any effect on the output? perhaps I am tired and being sloppy but I am not seeing any increase in granularity in the results

Will check

kshefchek:
I'm able to see the new relations in the orphanet output.

> @kshefchek Do you mean we should get rid of these from the local translation table if they aren't being used?

I see these in there now, must have missed them.  We should update their mappings to these RO terms:

Disease-causing germline mutation(s) (loss of function) in: RO:0004011
Disease-causing germline mutation(s) (gain of function) in: RO:0004012

then I think we can get it merged!

cmungall:
I assume this means some of the queries that build solr would be modified -
might be useful if someone shows Justin how this part works

justaddcoffee:
n there now, must have missed them. We should update their mappings to these RO terms:
> 
> Disease-causing germline mutation(s) (loss of function) in: RO:0004011
> Disease-causing germline mutation(s) (gain of function) in: RO:0004012
> 
> then I think we can get it merged!

Okay @kshefchek, just made these changes above per your request and pushed to this branch. Tests pass and dipper-etl.py -s orphanet completes without incident. Should be ready to merge...

justaddcoffee:
> I assume this means some of the queries that build solr would be modified - might be useful if someone shows Justin how this part works

Okay, I'd be interested in a tour of solr and how it might have changed related queries...

kshefchek:
+1, I'll merge since you may be blocked with travis failing

justaddcoffee:
Thanks @kshefchek

PR #725, opened 2019-04-16
OMIM cleanup

TomConlin:
Removed OMIM being subclass_of things which are not gene (or obsolete)
References to NCBIGenome are switched to  NCBIAssembly
Phenotype Series logic gone over

Refactor 
 - short variables to a three char min 
 - magic numbers to tt labels

Notes where the ingest exceeds its remit


dipper/sources/OMIM.py
```diff
@@ -471,58 +474,70 @@ def _transform_entry(self, e, graph, globaltt):
...
+                model.addClassToGraph(omim_curie, newlabel, omimtype)
             else:
-                model.addClassToGraph(omim_curie, newlabel, omimtype)  # Do subclass X?
+                # omim NOT subclass_of D|P|or ?...
+                model.addClassToGraph(omim_curie, newlabel)
 
             # add the original screaming-caps OMIM label as a synonym
-            model.addSynonym(omim_curie, label)
+            # P.R. says these are not an exact synonym in #714
+            model.addSynonym(omim_curie, label, model.globaltt['has_related_synonym'])
```
kshefchek: I don't think we need to add synonym data here, but it's not a big deal either way

dipper/sources/OMIM.py
```diff
@@ -471,58 +474,70 @@ def _transform_entry(self, e, graph, globaltt):
...
 
-            # add the alternate labels and includes as synonyms
+            # add the alternate labels and includes as synonym-ish
             for label in other_labels:
                 model.addSynonym(
                     omim_curie, label, model.globaltt['has_related_synonym'])
 
+            if abbrev is not None:
+                model.addSynonym(omim_curie, abbrev)  # as an exact synonym
```
kshefchek: Same as above

kshefchek:
Some comments on synonyms but looks good to me.

> References to NCBIGenome are switched to NCBIAssembly

I don't have a good enough understanding of this portion of dipper and how these connect to other sources, but likely will not break anything.

TomConlin:
The local_ID `GRCh38` appended is for a NCBI Assembly not a  NCBI genome.
 it is used to say what the bogus chromosome identifiers created are in reference to.
otherwise we would just be repeating  

```     <BOGUS> rdf:type <BOGUS> .```

which we do more than enough of with  `OBO:CHR_9606chr19`
which resolves to a nice big xml file describing Amazon S3 buckets.

```
roqet -q -r turtle -e "construct where{<https://monarchinitiative.org/MONARCH_GRCh38chr19> ?p ?o .}"  -D out/omim.ttl

<https://monarchinitiative.org/MONARCH_GRCh38chr19>
    <http://purl.obolibrary.org/obo/RO_0002350> <https://www.ncbi.nlm.nih.gov/assembly?term=GRCh38> ;
    a <http://purl.obolibrary.org/obo/CHR_9606chr19>, <http://purl.obolibrary.org/obo/SO_0000340> ;
    <http://www.w3.org/2000/01/rdf-schema#label> "chr19 (GRCh38)" .

```

File: dipper/sources/OMIMSource.py [hunk 3]
```diff
@@ -0,0 +1,223 @@
...
+                    self.omim_type[omim_id] = self.globaltt['gene']
+                elif declared == 'NULL':
+                    self.omim_type[omim_id] = self.globaltt['Suspected']   # NCIT:C71458
+                elif declared == 'Number Sign':
+                    self.omim_type[omim_id] = self.globaltt['Phenotype']
+                elif declared == 'Percent':
+                    self.omim_type[omim_id] = self.globaltt[
+                        'heritable_phenotypic_marker']
+                elif declared == 'Plus':
+                    self.omim_type[omim_id] = self.globaltt['has_affected_feature']
```
kshefchek: My understanding is that downstream this is treated like a gene, or are there cases where it is treated like other types?  I think typing a class/individual as a object property could be potentially confusing even if it doesn't make it's way to the RDF model.

TomConlin: which `this` are we referring to?

only omim identifiers of type 'gene' are being declared subclasses of the type
 as to what types to consider each of the choices I'm 100% open
but they each need a different type here.

kshefchek: Referring to:

    elif declared == 'Plus':
        self.omim_type[omim_id] = self.globaltt['has_affected_feature']

We've previously considered these genes, for example:
https://monarchinitiative.org/gene/OMIM:100650

TomConlin: I think we are ok,  these types are not binding in rdf-land
just a way to make as fine a distinction as we can. 
https://github.com/TomConlin/dipper/blob/master/dipper/sources/OMIM.py#L392

TomConlin:
# Update
This PR became larger than just OMIM in that OMIM ids are consulted in multiple ingests.
 the job of deciding what __type__ of OMIM identifier an ID is has beed moved to a super class
 and is shared by all the ingests which had been, either rolling their own or using an expensive but  dangerous utility function.

Other requests such as avoid emitting alternative names were implemented
and new conventions such as vetting column headers, more formally introduced  
 (have been inching it that direction for quite a while)   

As it stands the new output has ~100k fewer statements that previously
although slightly surprisingly it has nearly 1k more "Monarch_associations"
the vast majority of these have ncbi-gene identifiers as subjects which are already subjects of other "Monarch_associations.

There are also 
 - 45 Monarch_association subjects which are lost 
 - 40  Monarch_association subjects which are gained
all but two or three of these are also ncbi-gene identifiers

Spot checking these additions and deletions was unsatisfying.  
I could not see from the records at OMIM why they should be treated differently.
It may end up being something odd like the inclusion of an abbreviation in the json;  
but I feel I must get moving one other things at this point.

PR #730, opened 2019-04-25
Update GO gene annotation URLs

justaddcoffee:
Updated some GO annotation files. Should fix some errors in the current Dipper ingest. 

I've changed the domain to pull GO gene annotations from http://current.geneontology.org/annotations/ per Seth's suggestion. This current subdomain contains the "release" version of these files. These should be fairly stable.


TomConlin:
I like it. guaranteed not worse. 
Keeping line length, including comments, to not greater than 88 columns
will reduce the screeching of my static checker.
(note: the arbitrary length is with an eye towards eventually employing [black](https://github.com/python/black) on this codebase)

PR #732, opened 2019-05-06
Yaml Lint 

TomConlin:
Another basic test for the translation tables,   
this time validating them as compliant YAML.

`yamllint` should be available via your package manager but if not see:  
https://github.com/adrienverge/yamllint  

- call embedded within `Makefile` in rule for `make trans-test` which is emitting less cruft
- default tt files are written slightly better filled out as a template
- found and cleaned up a number of minor complaints
- led to including an explicit default in resolve() which may be helpful as a workaround when a source has smallish key conflicts. 
    - will get first use in MGI.py with "Not Applicable" & "Not Specified"
 - IMPC  what a mess!
    - fixing is beyond the scope of this PR but it is not worse that it was.
    - duplicate key with the same values are removed
    - duplicate keys with different value have arbitrary-tie-breaker-comment-out applied.
- there is a tiny `.yamllint` config file in the translation table directory   
 to suppress long line complaints

Just remember to install `yamllint` if need be before running.


justaddcoffee:
> Another basic test for the translation tables,
> this time validating them as compliant YAML.
> 
> `yamllint` should be available via your package manager but if not see:
> https://github.com/adrienverge/yamllint
> 
> * call embedded within `Makefile` in rule for `make trans-test` which is emitting less cruft
> * default tt files are written slightly better filled out as a template
> * found and cleaned up a number of minor complaints
> * led to including an explicit default in resolve() which may be helpful as a workaround when a source has smallish key conflicts.
>   
>   * will get first use in MGI.py with "Not Applicable" & "Not Specified"
> * IMPC  what a mess!
>   
>   * fixing is beyond the scope of this PR but it is not worse that it was.
>   * duplicate key with the same values are removed
>   * duplicate keys with different value have arbitrary-tie-breaker-comment-out applied.
> * there is a tiny `.yamllint` config file in the translation table directory
>   to suppress long line complaints
> 
> Just remember to install `yamllint` if need be before running.

I added yamllint to requirements.txt thusly
git+https://github.com/adrienverge/yamllint
in Tom's master branch, in order for yamllint to installed during regular installation process. Possibly brittle if adrienverge's git repo URL is going to be unstable

justaddcoffee:
Also added this to Makefile to disambiguate errors caused by malformed yaml and errors caused by yamllint not being installed or not found in path

	@ echo  "Is yamllint installed?"
	@ yamllint

TomConlin:
- whinging  good.  
 - non essential requirements bad.

(please notice I have been going out of my way to erode requirements whenever possible)

PR #733, opened 2019-05-08
GWAS Catalog

TomConlin:
### major changes
  - standard file header checks
  - new SO terms in global tt
  - blank nodes get proper opaque identifiers and words are attached with labels
    - note: want to see this behaviour everywhere (and bnode types eventually)
  - reduced logging output that was uninformative (being verbose about empty string)
  - find a significant number of compound SO terms (crosses)
    - they are commented out in the local tt file. welcome thoughts on what to do with them

 
fix typo in Jenkins file


dipper/sources/GWASCatalog.py
```diff
@@ -263,69 +318,73 @@ def _process_haplotype(
...
             graph.addTriple(hap_id, self.globaltt['has_variant_part'], snp_curie)
             snp_curies.append(snp_curie)
 
         # courtesy http://stackoverflow.com/a/16720915
-        length = len(snp_labels)
+        # check lengths of mutiple lists
+        length = len(snp_curies)
         if not all(len(lst) == length
-                   for lst in [chrom_nums, chrom_positions, context_list]):
+                   for lst in [snp_labels, chrom_nums, chrom_positions, context_list]):
```
kshefchek: This is evaluating as False when the condition is true, change to:

    if not all([len(lst) == length
               for lst in [snp_labels, chrom_nums, chrom_positions, context_list]]):

dipper/sources/GWASCatalog.py
```diff
@@ -263,69 +318,73 @@ def _process_haplotype(
...
-        if len(mapped_genes) == variant_in_gene_count and len(set(mapped_genes)) == 1:
-            gene_id = DipperUtil.get_hgnc_id_from_symbol(mapped_genes[0])
-            geno.addAffectedLocus(hap_id, gene_id)
+                    gene_id = DipperUtil.get_hgnc_id_from_symbol(mapped_genes[index])
+                    if gene_id is not None \
+                            and context_list[index] in ['upstream_gene_variant',
+                                                        'downstream_gene_variant']:
+                            graph.addTriple(
+                                snp_curie, self.resolve(context_list[index]), gene_id)
+                    else:
```
kshefchek: Over indented, I think this relates to the preceding if len(mapped_genes) == ...

kshefchek:
Please fix test/test_gwascatalog.py (even if it's updating the bnode IRIs)

PR #735, opened 2019-05-09
Update fix zfin ingest

justaddcoffee:
To address https://github.com/monarch-initiative/dipper/issues/723


dipper/sources/ZFIN.py
```diff
@@ -2495,19 +2439,19 @@ def process_fish_disease_models(self, limit=None):
         return
 
     def _map_sextuple_to_phenotype(
-            self, superterm1_id, subterm1_id, quality_id, superterm2_id,
-            subterm2_id, modifier):
+            self,
+            superterm1_id, subterm1_id, quality_id,
```
matentzn: This will probably work, but maybe use the relationship columns as well.

justaddcoffee: I actually did use the relationship columns in a [previous commit on this branch](https://github.com/monarch-initiative/dipper/pull/735/commits/3b621c278caf8312bece884651cc9c3caaf565d9#diff-57a8cd3fe591cdb53eebc04f55bc8295L2497)
Could roll back and do that, but my concern is that if we add these two relationships columns we will lose some g2p triples that are present in the current ingest if even a single one of the relationship ids doesn't match up to what's in phenotype_fish.txt or phenoGeneCleanData_fish.txt

dipper/sources/ZFIN.py
```diff
@@ -17,7 +17,7 @@
 
 LOG = logging.getLogger(__name__)
 ZFDL = 'http://zfin.org/downloads'
-
+OBO = 'http://purl.obolibrary.org/obo/zp/src/curation'
```
TomConlin: OBO is defined in curie map, it would be good to choose a non conflicting variable name so when we come across it in any dipper code we have a good sense of what it is

dipper/sources/ZFIN.py
```diff
@@ -1247,16 +1246,15 @@ def _process_g2p(self, limit=None):
         with open(raw, 'r', encoding="utf8") as csvfile:
             filereader = csv.reader(csvfile, delimiter='\t', quotechar='\"')
             for row in filereader:
-                line_counter += 1
-
+                line_counter += 1 
+
```
TomConlin: I have been eliminating these counters as I go and using the one builtin and maintained by the ```csv.reader``` .
so later when "line_counter" would be used, instead favor `filereader.line_num`.
To help keep that reasonable I have also been choosing `reader'`  `writer` over 
"strangely_specific_filesystem_input-output_handler"  or "fh" or "f"

tests/test_zfin.py
```diff
@@ -36,5 +36,18 @@ def test_allZPAvailable(self):
...
         return
 
+    def test_load_zp_mappings(self):
+        if self.source is not None:
+            try:
+                zp_map = self.source._load_zp_mappings("./tests/resources/zfin/zp-mapping-test.txt")
+                self.assertIsInstance(zp_map, dict, "_load_zp_mappings() didn't return dict!")
+                self.assertTrue(len(zp_map) == 1, "_load_zp_mappings() didn't return exactly one thing!")
+                self.assertDictEqual(zp_map, {'MONARCH:b311b709bf1cf852d770': {'zp_id': 'ZP:0002959', 'subterm1_id': 'ZFA:0009114', 'post_composed_relationship_id_1': 'BFO:0000050', 'superterm1_id': 'ZFA:0001056', 'quality_id': 'PATO:0001453', 'subterm2_id': 'GO:0005927', 'post_composed_relationship_id_2': 'BFO:0000050', 'superterm2_id': 'ZFA:0001056'}}, "_load_zp_mappings() didn't return what I expected!")
+            except Exception as e:
```
TomConlin: Going to ask that __all__ new variables have a minimum of three characters  
unless contained on one line, such as within a list comprehension. 
(replacing existing single char vars as you go is also highly appreciated)

dipper/sources/ZFIN.py
justaddcoffee: Thanks Tom! Done.

tests/test_zfin.py
justaddcoffee: Thanks, done, and I agree - will replace these going forward if/when I come across them

dipper/sources/ZFIN.py
justaddcoffee: Sensible, and will do this in ZFIN.py as soon as I can. Appreciate the style tips Tom!

dipper/sources/ZFIN.py
matentzn: Yes, but this is exactly the reason they exist: they should not match up. Basically, if it is not in the map, it should not be ingested; so rather than allowing here to map to incomplete records (without the relationships), we need to continue to work expanding the coverage of these mappings in the ZP further..

justaddcoffee: Ah okay - duly noted, thanks Nico

TomConlin:
@justaddcoffee  what is the merge plan here?

justaddcoffee:
@TomConlin nearly there, need to update ZFINSlim.py and then we can probably merge. I'll try to sort out ZFINSlim.py today-ish

justaddcoffee:
> @justaddcoffee what is the merge plan here?

Barring any objection from @matentzn, I think we can merge this. I'd like to include these changes in the latest data release that @TomConlin is working on

dipper/sources/ZFINSlim.py
```diff
@@ -19,13 +19,13 @@ class ZFINSlim(Source):
...
         'g2p_clean': {
-            'file': 'phenoGeneCleanData_fish.txt.txt',
+            'file': 'phenoGeneCleanData_fish.txt',
             'url': 'https://zfin.org/downloads/phenoGeneCleanData_fish.txt',
             # https://zfin.org/downloads#  header Documentation is burried in UI crap
         },
         'zpmap': {
-            'file': 'zp-mapping.txt',
-            'url': 'http://compbio.charite.de/hudson/job/zp-owl-new/lastSuccessfulBuild/artifact/zp.annot_sourceinfo'
+            'file': 'zp-mapping-2019.txt',
```
matentzn: Why is the year in here? Will this not be superseded very quickly?

dipper/sources/ZFINSlim.py
```diff
@@ -81,12 +81,13 @@ def parse(self, limit=None):
...
                  pub_id,
                  figure_id
-                 # ,unknown_field   # just leave this here for next time
                 ) = row
 
-                zp_id = zfin_parser._map_sextuple_to_phenotype(
-                    superterm1_id, subterm1_id, quality_id, superterm2_id,
-                    subterm2_id, modifier)
+                if ( modifier != "abnormal"):
+                    LOG.warning("skipping phenotype with modifier != abnormal: " + modifier)
```
matentzn: Great!

tests/test_zfin.py
```diff
@@ -21,20 +21,53 @@ def tearDown(self):
...
+            'ZP:0022140')
         return
 
+    def test_load_zp_mappings(self):
+        if self.source is not None:
+            try:
+                zp_map = self.source._load_zp_mappings("./tests/resources/zfin/zp-mapping-test.txt")
+                self.assertIsInstance(zp_map, dict, "_load_zp_mappings() didn't return dict!")
+                self.assertTrue(len(zp_map) == 1, "_load_zp_mappings() didn't return exactly one thing!")
+                self.assertDictEqual(zp_map, {'MONARCH:b308a8f1c67793a56d16': {'post_composed_relationship_id_1': 'BFO:0000050',
```
matentzn: Nice.

translationtable/zfinslim.yaml
```diff
@@ -0,0 +1,3 @@
+---
+# zfinslim.yaml
+"": ""  # example
```
matentzn: Purpose?

dipper/sources/ZFIN.py
```diff
@@ -17,7 +19,7 @@
 
 LOG = logging.getLogger(__name__)
 ZFDL = 'http://zfin.org/downloads'
-
+OBOLIB = 'http://purl.obolibrary.org/obo/zp/src/curation'
```
matentzn: Variable name is misleading: Better use: **_ZPCURATION_**

dipper/sources/ZFIN.py
```diff
@@ -2494,46 +2324,58 @@ def process_fish_disease_models(self, limit=None):
...
+        :param post_composed_relationship_id_2,
+        :param superterm2_id,
+        :param modifier
         """
+
         zp_id = None
 
         # zfin uses free-text modifiers,
         # but we need to convert them to proper PATO classes for the mapping
         mod_id = self.resolve(modifier, False)
```
matentzn: I am not sure why the actual PATO modifier is needed at all here. Why does not the string 'abnormal' suffice? I am only asking because I would like to be sure that mod_id is not used anywhere for anything else other than resolving the mapping.. (especially not added to the knowledge graph).

dipper/sources/ZFIN.py
```diff
@@ -2544,43 +2386,54 @@ def _load_zp_mappings(self, file):
...
         :return:
+        dict zp_map with 
 
         """
         zp_map = {}
         LOG.info("Loading ZP-to-EQ mappings")
         line_counter = 0
         with open(file, 'r', encoding="utf-8") as csvfile:
             filereader = csv.reader(csvfile, delimiter='\t', quotechar='\"')
+            header_or_comment_re = re.compile('^iri|^#')
```
matentzn: Nice, to allow comments!

dipper/sources/ZFIN.py
```diff
@@ -2544,43 +2386,54 @@ def _load_zp_mappings(self, file):
...
             for row in filereader:
+                if header_or_comment_re.match(row[0]):
+                    continue
                 line_counter += 1
-                (zp_id, zp_label, superterm1_id, subterm1_id, quality_id,
-                 modifier, superterm2_id, subterm2_id) = row
+
+                zp_id = row[0]
+                modifier = self.resolve("abnormal", False) # Nico assures us that id_map_zfin.tsv only contains data for abnormal phenotypes
```
matentzn: This is true, but it is, in theory, possible that one day, normal phenotype statements are allowed. Lets be sure that:
1. We filter out everything that is not abnormal
2. We never make the assumption that anything we find in a ZFIN dataset IS in fact abnormal

Both must be true.

dipper/sources/ZFINSlim.py
justaddcoffee: Basically to stop from using the old files while I was testing, and also to indicate when the mapping was made. Easy enough to change

translationtable/zfinslim.yaml
justaddcoffee: This was autogenerated by dipper-etl.py. Not sure why it made this file, but it probably needs to be in the repo

dipper/sources/ZFIN.py
justaddcoffee: Agreed

dipper/sources/ZFIN.py
justaddcoffee: generally the way things work in Dipper is to look things like this ("abnormal" in this case) in the translation table using .resolve, and match things using ontology ids where possible. this prevents things like "abnormal" not matching "abnormal " or "Abnormal" or " abnormal ", among other advantages.

dipper/sources/ZFIN.py
justaddcoffee: I aim to please

dipper/sources/ZFIN.py
justaddcoffee: Ah okay, I'll take this out - this was just a note I made to myself while we were meeting. It does seem a little hostile now that I read it! Sorry about that!

dipper/sources/ZFINSlim.py
matentzn: sure doesnt matter.

dipper/sources/ZFIN.py
matentzn: Ah nooo! This was not what I read. :P Like you, I just want the ingest to be future-proof, thats all. :P

justaddcoffee:
hey @TomConlin I'm going to meet with @matentzn about this PR on Friday. Is this too late to make it into the next data release? I'd really like this new ZFIN ingest to go in there...

dipper/sources/ZFIN.py
justaddcoffee: we possibly then should make an item in your file explicitly stating "abnormal" or PATO:0000460 (even though every item will have it), and then pass out those that don't, to help prevent normal phenotypes from slipping into the ingest in the future...

matentzn: Hmmmmmmm... I think I have to admit you might be right. Can you do the following: when you parse the mapping file, after the split, if there is no 8th element in the split, assume it is PATO: 0000460, and when there is an 8th element, take that instead. That way we dont have to wait until I actually add the additional column, but if I do, and something other than PATO: 0000460 will be added, it will be robust and work.

matentzn: For now though: leave the abnormal filter in there. Just add an assumption that if there are only 7 elements in the split, the 8th is PATO: 0000460.

matentzn:
@TomConlin What is your exact time and date for the ingest? :P We will make it for that no matter what!

dipper/sources/ZFIN.py
justaddcoffee: we could do that, although it makes the code and zp-mapping-2019.txt a little more opaque and confusing for the next person who has to maintain the code...

justaddcoffee: Actually I think it might also be a good idea to put a header in the zp-mapping file also explaining what each field is (also in the interest of making the code/data a bit more intuitive), but that's not a show stopper

TomConlin:
@matentzn  what is this "exact" of which you speak?

matentzn:
It’s a phrase we use in Germany sometimes to confuse the people we work with :P if Friday is good: awesome! Let’s talk then. 8.30 am your time

justaddcoffee:
@matentzn okay, talk then. @TomConlin feel free to join if you can't get enough of ZFIN

justaddcoffee:
per your request @matentzn, here are the deltas of each predicate between this zfin ingest and the Feb 2019 one. I'm posting this here, b/c I wanted to see what @TomConlin @kshefchek thought about my writing some software to automatically compare the [bleeding edge ingests here](https://data.monarchinitiative.org/ttl/), with [all previous ingests](https://data.monarchinitiative.org/previous/), and automatically make some graphs and tables to help more easily analyze what's changed between ingests: for each ingest, total number of triples, total number of each predicate, change in predicate count, etc

[zfinslim_201901.ttl_vs_zfinslim_update_fix_zfin_ingest.ttl.tsv.txt](https://github.com/monarch-initiative/dipper/files/3234735/zfinslim_201901.ttl_vs_zfinslim_update_fix_zfin_ingest.ttl.tsv.txt)
[zfin_201901.ttl_vs_zfin_update_fix_zfin_ingest.ttl.tsv.txt](https://github.com/monarch-initiative/dipper/files/3234736/zfin_201901.ttl_vs_zfin_update_fix_zfin_ingest.ttl.tsv.txt)

matentzn:
@justaddcoffee Man this is really cool! Thanks! This help, I think, a lot with debugging! Based on this code, you can write a test that sends a warning email when there where changes +-50% in any of the relations, so that someone takes a look; this will hopefully help with the problem of schema changes on the source side.. 16% increase in has phenotype relations sounds right! Just one question: The increase in http://www.w3.org/2002/07/owl#equivalentClass, how did that happen? Is this involve blank nodes (hopefully)? There should not be any http://www.w3.org/2002/07/owl#equivalentClass ingested between names.

justaddcoffee:
@TomConlin ready to merge this PR whenever you are

dipper/sources/ZFIN.py
```diff
@@ -1247,16 +1134,15 @@ def _process_g2p(self, limit=None):
         with open(raw, 'r', encoding="utf8") as csvfile:
             filereader = csv.reader(csvfile, delimiter='\t', quotechar='\"')
             for row in filereader:
-                line_counter += 1
-
+                line_counter += 1 
+
```
TomConlin: trailing spaces  will cause incessant nagging and annoying diffs

justaddcoffee: Okay, sorry about that, pylint says no trailing whitespace now

dipper/sources/ZFIN.py
```diff
@@ -17,7 +19,7 @@
 
 LOG = logging.getLogger(__name__)
 ZFDL = 'http://zfin.org/downloads'
-
+ZPCURATION = 'http://purl.obolibrary.org/obo/zp/src/curation'
```
TomConlin: two blank lines before a class def

dipper/sources/ZFIN.py
```diff
@@ -69,8 +71,8 @@ class ZFIN(Source):
             'file': 'zfinpubs.txt',
             'url': ZFDL + '/zfinpubs.txt'},
         'zpmap': {
-            'file': 'zp-mapping.txt',
-            'url': 'http://compbio.charite.de/hudson/job/zp-owl-new/lastSuccessfulBuild/artifact/zp.annot_sourceinfo'},
+            'file': 'zp-mapping-2019.txt',
+            'url': ZPCURATION + '/id_map_zfin.tsv'}, # this is a zp mapping file generated by Nico
```
TomConlin: two spaces before an inline comment
but this one exceeds 88 col so move to proceeding line.

dipper/sources/ZFIN.py
```diff
@@ -2544,43 +2394,64 @@ def _load_zp_mappings(self, file):
...
-            superterm2_id, subterm2_id, modifier):
-        key = self.make_id('_'.join((
-            superterm1_id, subterm1_id, quality_id,
-            superterm2_id, subterm2_id, modifier)))
+            subterm1_id, post_composed_relationship_id_1, superterm1_id,
+            quality_id, subterm2_id, post_composed_relationship_id_2,
+            superterm2_id,
+            modifier,
+            replace_empty_args_with_zeros=True):
+    # empty args -> zero is the convention in Nico's file and elsewhere
```
TomConlin: mis indented comment

tests/resources/zfin/zfin_test_ids.yaml
```diff
@@ -0,0 +1,575 @@
+genotype:
```
TomConlin: as mentioned in comments last week;
This file is formatted differently than the existing test_id.yaml files.
for example please see:
https://github.com/monarch-initiative/dipper/blob/master/resources/test_ids.yaml

dipper/sources/ZFIN.py
justaddcoffee: thanks, fixed

dipper/sources/ZFIN.py
justaddcoffee: fixed

dipper/sources/ZFIN.py
justaddcoffee: fixed

tests/resources/zfin/zfin_test_ids.yaml
justaddcoffee: Hmm - [that file](https://github.com/monarch-initiative/dipper/blob/master/resources/test_ids.yaml) doesn't seem to be valid YAML...

TomConlin: there are lines longer than 80 cols. (see the Makefile for how we ignore them)

and there are two commas that need a space after them...

good catch

PR #736, opened 2019-05-10
Remove maven directory

kshefchek:
Once https://github.com/monarch-initiative/scigraph-docker is fully tested, we should remove the maven directory used for the legacy docker pipeline


TomConlin:
wahoo!

kshefchek:
apparently this is in the solr docker build as well, so closing for now until that dockerfile is cleaned up, https://github.com/monarch-initiative/solr-docker-monarch-golr/blob/master/Dockerfile#L52

PR #757, opened 2019-05-28
added simple method to compare counts of predicates between two graphs

justaddcoffee:
needed this for two things at the moment, so thought it might be generally useful

### DIFF
--- a/tests/test_graph_utils.py
+++ b/tests/test_graph_utils.py
@@ -6,7 +6,7 @@ import rdflib
 from dipper.utils import GraphUtils
 
 logging.basicConfig(level=logging.WARNING)
-logger = logging.getLogger(__name__)
+LOG = logging.getLogger(__name__)
 
 class GraphUtilsTestCase(unittest.TestCase):
 
@@ -20,7 +20,6 @@ class GraphUtilsTestCase(unittest.TestCase):
         self.test_graph2_id = self.test_graph2.identifier
         self.enemyOf = rdflib.term.URIRef('http://www.perceive.net/schemas/relationship/enemyOf')
         self.name = rdflib.term.URIRef('http://xmlns.com/foaf/0.1/name')
-        return
 
     def tearDown(self):
         return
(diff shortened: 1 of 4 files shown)


kshefchek:
+1

tests/test_graph_utils.py
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+
+import unittest
+import logging
+import rdflib
+from dipper.utils import GraphUtils
+
+logging.basicConfig(level=logging.WARNING)
+logger = logging.getLogger(__name__)
```
TomConlin: other ingest all use `LOG = `  because it is short and a GLOBAL

tests/test_graph_utils.py
```diff
@@ -0,0 +1,52 @@
...
+        self.graph_object = rdflib.Graph()
+        self.graph_object2 = rdflib.Graph()        
+        self.graph_util = GraphUtils.GraphUtils
+        self.test_graph = self.graph_object.parse("tests/resources/graphutils/gu_test_graph.ttl", format="ttl")
+        self.test_graph_id = self.test_graph.identifier
+        self.test_graph2 = self.graph_object2.parse("tests/resources/graphutils/gu_test_graph2.ttl", format="ttl")
+        self.test_graph2_id = self.test_graph2.identifier
+        self.enemyOf = rdflib.term.URIRef('http://www.perceive.net/schemas/relationship/enemyOf')
+        self.name = rdflib.term.URIRef('http://xmlns.com/foaf/0.1/name')
+        return
```
TomConlin: this is a point of no `return`

tests/test_graph_utils.py
justaddcoffee: fixed

tests/test_graph_utils.py
justaddcoffee: fixed

TomConlin:
dipper/utils/GraphUtils.py 

    standard imports come before required or local imports
    trailing whitespace
    lines <= 88 cols (leftmost continuation preferred)
    one blank line between functions (two before class)
    omit empty return

justaddcoffee:
okay @TomConlin think I've linted those issues

Issue #766, opened 2019-06-14
Latest GO ingest missing human data


kshefchek:
Comparing 
http://data.monarchinitiative.org/ttl/go.ttl
to
https://archive.monarchinitiative.org/201902/ttl/go.ttl

It looks like we've lost some data.  We no longer have any NCBIGene or  ENSEMBL curies so it's possible the drop is coming from the human data.

TomConlin:
Have re factored, 
included two new taxon mould & fungi

It is now returning NCBIGene and ENSEMBL curies. 

It is casting some curies as literals via `has_exact_synonym`  will try to address

PR #767, opened 2019-06-18
Go annotation

TomConlin:
addresses #766 

refactor column headers etc
add s couple of new species types(mould & fungi)
bump date for monarch archive
change  

    '<gene> <has_synonym> "prot_curie" .'
to    

     '<gene> <same _as> <prot_curie> .'

(open to alternative associations)

### DIFF
--- a/dipper/sources/GeneOntology.py
+++ b/dipper/sources/GeneOntology.py
@@ -121,11 +121,22 @@ class GeneOntology(Source):
             'url': GOGA + '/sgd.gaf.gz',
             'columnns': gaf_columns
         },
-        '4896': {
+        '4896': {  # Schizosaccharomyces pombe  (yeast)
             'file': 'pombase.gaf.gz',
             'url': GOGA + '/pombase.gaf.gz',
             'columnns': gaf_columns
         },
+        '5782': {  # Dictyostelium (slime mold)
+            'file': 'dictibase.gaf.gz',
+            'url': GOGA + '/dictybase.gaf.gz',
+            'columnns': gaf_columns
+        },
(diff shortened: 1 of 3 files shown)


kshefchek:
> '<gene> <same _as> <prot_curie> .'

I don't mind the simplification but we probably should align each source with this change.  In other sources we use 'has gene product' to link proteins to genes.  So if we half commit to gene-protein equivalence we'll end up with has gene product cycles.

dipper/sources/GeneOntology.py
```diff
@@ -318,46 +387,46 @@ def process_gaf(self, file, limit, id_map=None, eco_map=None):
...
                 # Derive G2P Associations from IMP annotations
                 # in version 2.1 Pipe will indicate 'OR'
                 # and Comma will indicate 'AND'.
                 # in version 2.0, multiple values are separated by pipes
                 # where the pipe has been used to mean 'AND'
                 if eco_symbol == 'IMP' and with_or_from != '':
-                    withitems = re.split(r'\|', with_or_from)
-                    phenotypeid = go_id+'PHENOTYPE'
+                    withitems = with_or_from.split('|')
+                    phenotypeid = go_id + 'PHENOTYPE'
```
kshefchek: I know it's just formatting, but should we be moving away from these? cc @matentzn @cmungall

TomConlin:
I like 'has gene product' much better than 'same_as'  
which a grabbed base on being proximally the closest thing to has_synonym in the model class

dipper/sources/GeneOntology.py
TomConlin: +1

PR #768, opened 2019-06-18
Gene Ontology

TomConlin:
Review should include previous PR  #767 which I pushed to test a full run in the pipeline.
see: https://github.com/monarch-initiative/dipper/pull/767

mostly a refactor  and to address missing gene_id mappings from uniprot  see #766
now includes a couple of more primitive taxon (mould & fungi)
avoids emitting known curies as literal synonyms 
now logs unhanded curies
improved comments
make cached ID file names not dependent on command line argument order
make count of missing mappings (may address in another PR)


kshefchek:
+1

Issue #752, opened 2019-05-17
Remove blank node "models" in flybase and wormbase


kshefchek:
Currently we create blank nodes to attach some entity to a disease via the "is model of" relation.  We want to remove these and connect which ever the source asserts to be the model of the disease (gene, variant, genotype).

See https://github.com/monarch-initiative/dipper/issues/506

kshefchek:
Blank node models removed from Flybase in https://github.com/monarch-initiative/dipper/pull/760

PR #772, opened 2019-06-22
Simplifying WormBase modeling

kshefchek:
1. Removed test_mode code, wrote shell script to generate test set data
2. Removed blank node models, diseases are now connected to genes
3. Removed blank node alleles, WBVars are now being treated like alleles and sequence alterations
4. Remove blank node reagent targeted genes, phenotypes are now connected to the WBRNAi ids 
(Reagent targeted gene - is expression variant of -> gene_id)
(Reagent targeted gene - has phenotype -> phenotype), see https://www.wormbase.org/species/c_elegans/rnai/WBRNAi00080586

fixes https://github.com/monarch-initiative/dipper/issues/752


kshefchek:
As a final pass I've changed the typing of RNAi ids to be reagent targeted genes instead of RNAi reagents, since it makes more sense to connect phenotypes to RTGs instead of reagents.  We could type the IRI as both, but I'm not sure we gain much from this.

PR #778, opened 2019-06-28
minor fixes to GLOBAL_TERMS.yaml

justaddcoffee:
fixed incorrect SO id for sense_intronic_lncRNA_gene, and fixed ordering of sense_overlap_ncRNA_gene and sense_intronic_ncRNA_gene


justaddcoffee:
@TomConlin add in any other little fixes you can think of

PR #783, opened 2019-07-05
OMIM phenotypic series inclusion

TomConlin:
bring back phenotypic series.
changing a test from not greater than one to exactly one earlier this year 
unnecessary caused a drop in the number of PhenotypicSeries to OMIM associations.
the solution is to remove the dependency altogether (including when test is greater than one)


TomConlin:
Travis succeeding with python3.5 and failing with python3.6  while loading `requirements.txt` has nothing to do with this P.R.

PR #788, opened 2019-07-13
DRY out jenkins pipeline; add gard, cohd etc to curie map

kshefchek:
Jenkins pipeline
- Build dipper once
- Move files in single step
- Add ucscbands and omia

Curie map:
- Add gard, cohd, sctid, icd9


kshefchek:
im going to cowboy this in given there are no changes to the codebase, we can revert need be

TomConlin:
retroactive +1

PR #790, opened 2019-07-16
Clinvar assert RDF object is not None

TomConlin:
Linting
  - lines longer than 88
  - wrong number of blank lines

Guard against emitting `None`  as a disease label 

Started as an exercise to remove errant intermediate file
but the file only still existed due to the previous error.


kshefchek:
+1

PR #791, opened 2019-07-17
Ensembl refactor to fix missing column failure

TomConlin:
- @kshefchek so sorry I think my editor must have not liked whitespace changes in Jenkinsfile
 - @justaddcoffee this will take a careful merge with your injecting category edits

In general I am not fond of dipper-etl.py replacing arguments passed to sources


dipper/sources/Ensembl.py
```diff
@@ -129,69 +128,68 @@ def parse(self, limit=None):
...
+    #    header = next(response).rstrip().split('\t')
+    #    for line in response:
+    #        line = line.decode('utf-8').rstrip()
+    #        row = line.split('\t')
+    #        if len(row) != len(att):
+    #            LOG.warning("Data error for p-list query on %d", taxon_id)
+    #            continue
+    #        protein_list.append(row[att.index('ensembl_peptide_id')])
+    #    conn.close()
+    #    return protein_list
```
kshefchek: at one point using this in string, but we've probably moved past it, +1 for deleting unused code

TomConlin:
- python 3.6  passes 
 - python 3.5 fails with a complain about invalid syntax in RDG which is untouched
 - locally the Makefile fails with a " No rule to make target"

kshefchek:
> python 3.5 fails with a complain about invalid syntax in RDG which is untouched

Looks like ontobio has some code that uses variable type annotations which are new in 3.6.  There may be backports, or we could remove support for 3.5

cmungall:
Any reason to still support 3.5?

kshefchek:
@cmungall see https://github.com/monarch-initiative/dipper/pull/794

PR #793, opened 2019-07-18
OMIM, Monochrom/UCSCBands

TomConlin:
indel s in phenotypic series url


kshefchek:
Tests passing with 3.6, +1

PR #796, opened 2019-07-19
Remove testsuite from zfin

kshefchek:
Removes testsuite code that resulted in a handful of source filepaths being overwritten in dipper-etl.py.  Our qc caught this via:

Genotype Phenotype
Danio rerio | 653 (-101276)

Variant Phenotype
Danio rerio | 8132 (-83923)

Variant Genotype
Danio rerio | 4 (-153358)


TomConlin:
+1

PR #797, opened 2019-07-19
Post process IRI sweep

TomConlin:
cleaning up IRI for dot qc checks
it is nice to see old exceptions in ntriples2dot go away


kshefchek:
+1

PR #801, opened 2019-07-23
String Ensembl

TomConlin:
Move list creation for dict values from Ensembl  (where no list are seen yet)
to String where lists are appropriate for consistency with other conditions.

Ensembl: Write out intermediate input from query so we can know and record what we got.
(should likely do more of this as opportunity allows) 

String; refactor dict values update to be more pythonic


kshefchek:
+1

PR #808, opened 2019-08-06
Use wget to fetch panther files

kshefchek:
Our panther fetcher is not downloading complete files from the panther ftp (silently), this is a quick patch to use wget instead.


TomConlin:
yep

Issue #792, opened 2019-07-17
Fixes required for datasets.ttl


cmungall:
Currently the dataset.ttl we produce has some issues. Example:

```
@prefix : <https://monarchinitiative.org/> .
@prefix MonarchArchive: <https://archive.monarchinitiative.org/201907/> .
@prefix OBO: <http://purl.obolibrary.org/obo/> .
@prefix dcat: <http://www.w3.org/ns/dcat#> .
@prefix dcterms: <http://purl.org/dc/terms/> .
@prefix dctypes: <http://purl.org/dc/dcmitype/> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .
@prefix pav: <http://purl.org/pav/> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix xml: <http://www.w3.org/XML/1998/namespace> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

<https://archive.monarchinitiative.org/201907/ttl/ctd.ttl> a dctypes:Dataset ;
    dcterms:identifier "MonarchArchive:ttl/ctd.ttl" ;
    dcterms:issued "2019-07-16",
        "Wed-Jun-26-14-47-40-EDT-2019",
        "Wed-Jun-26-14-49-06-EDT-2019",
        "Wed-Jun-26-15-26-21-EDT-2019" ;
    dcterms:rights "http://ctdbase.org/about/legal.jsp" ;
    dcterms:title "Comparative Toxicogenomics Database" ;
    dcat:accessURL <http://ctdbase.org/reports/CTD_chemicals_diseases.tsv.gz>,
        <http://ctdbase.org/reports/CTD_genes_diseases.tsv.gz>,
        <http://ctdbase.org/reports/CTD_genes_pathways.tsv.gz> ;
    foaf:page <http://ctdbase.org> .

<https://archive.monarchinitiative.org/201907/ttl/ctd.ttlWed-Jun-26-14-47-40-EDT-2019> dcterms:isVersionOf "MonarchArchive:ttl/ctd.ttl" ;
    pav:version "Wed-Jun-26-14-47-40-EDT-2019" .

[...]

Issue #753, opened 2019-05-23
Clarify semantics of Version


cmungall:
We are working on https://github.com/monarch-initiative/monarch-ui/issues/58

We now have version showed on the data sources page, but it's not clear what it means. Source version or version of monarch ingest?

From this it seems like:

https://github.com/monarch-initiative/dipper/blob/b2c20cdcd60cfd4b482a7fed36ec381edf74d7c7/dipper/sources/Source.py#L198-L200

source version is used UNLESS not present THEN ingest version/date is used?

justaddcoffee:
addressed by #809

Issue #545, opened 2017-11-13
Dataset URIs in metadata and in main tll dumps are not connected


cmungall:
in panther.ttl

```
MonarchData:panther.ttl a owl:Ontology ;
    owl:versionIRI MonarchArchive:panther-2017-02-10-13-07.ttl ;
    owl:versionInfo "panther-2017-02-10-13-07" .
```

in panther_dataset.ttl:

```
@prefix : <https://monarchinitiative.org/> .

<https://monarchinitiative.org/2017-10-04T19:13:08> dct:isVersionOf :panther2017-10-05 ;
    dct:issued "2017-10-04T19:13:08"^^xsd:dateTime ;
    pav:version "2017-10-04T19:13:08" .

:panther a dctypes:Dataset ;
    dct:identifier "panther" ;
    dct:issued "2017-10-05" ;
    dct:license <http://www.pantherdb.org/terms/disclaimer.jsp> ;
    dct:title "Protein ANalysis THrough Evolutionary Relationships" ;
    dcat:accessURL <ftp://ftp.pantherdb.org/ortholog/current_release/Orthologs_HCOP.tar.gz>,
        <ftp://ftp.pantherdb.org/ortholog/current_release/RefGenomeOrthologs.tar.gz> ;
    foaf:page <http://pantherdb.org/> .

:panther2017-10-05 dct:isVersionOf :panther ;
    pav:version "2017-10-05" .
```

note that these 2 graphs don't link up! We need a triple connecting MonarchData:panther.ttl to MonarchData:panther, stating that the former is some formatted version of the latter logical resource.

Related: when we store this in Solr, we use our MonarchData:panther.ttl as the source. The actual linkage to the original source is not AFAICT apparent. I think we should use MonarchData:panther

TomConlin:
Definitely agree about not hard coding  the extension. just about to make a PR that makes generating non turtle files easier and safer and includes nquads.  And likewise I want the IRI for the named graph in the quads to be the IRI for the `foo_dataset.ttl` resource.

justaddcoffee:
addressed by #809

Issue #428, opened 2017-02-28
Documentation on source-URLs


drseb:
I would really like to have a way to explore which files (url) are actually put into dipper.

TomConlin:
A a quick and dirty zeroth approximation:

```
ack-grep "'url':" dipper/sources/*.py
dipper/sources/AnimalQTLdb.py
55:            'url': AQDL + '/tmp/QTL_Btau_4.6.gff.txt.gz'},
59:        #   'url': AQDL + '/tmp/QTL_UMD_3.1.gff.txt.gz'},
62:            'url': AQDL + '/export/KSUI8GFHOT6/cattle_QTLdata.txt'},
65:            'url': AQDL + '/tmp/QTL_GG_4.0.gff.txt.gz'},
68:            'url': AQDL + '/export/KSUI8GFHOT6/chicken_QTLdata.txt'},
71:            'url': AQDL + '/tmp/QTL_SS_10.2.gff.txt.gz'},
74:            'url': AQDL + '/export/KSUI8GFHOT6/pig_QTLdata.txt'},
77:            'url': AQDL + '/tmp/QTL_OAR_3.1.gff.txt.gz'},
80:            'url': AQDL + '/export/KSUI8GFHOT6/sheep_QTLdata.txt'},
83:            'url': AQDL + '/tmp/QTL_EquCab2.0.gff.txt.gz'},
86:            'url': AQDL + '/export/KSUI8GFHOT6/horse_QTLdata.txt'},
89:            'url': AQDL + '/export/KSUI8GFHOT6/rainbow_trout_QTLdata.txt'},
93:            'url': AQDL + '/export/trait_mappings.csv'}

dipper/sources/BioGrid.py
31:            'url': BGDL + '/BIOGRID-ALL-LATEST.mitab.zip'},
34:            'url': BGDL + '/BIOGRID-IDENTIFIERS-LATEST.tab.zip'}

dipper/sources/ClinVar.py
33:            'url': CVDL + '/variant_summary.txt.gz'
37:            'url':  CVDL + '/var_citations.txt'

dipper/sources/ClinVarXML_alpha.py
51:        'url': 'ftp://ftp.ncbi.nlm.nih.gov/pub/clinvar/xml/ClinVarFullRelease_00-latest.xml.gz'}

dipper/sources/CTD.py
64:            'url': 'http://ctdbase.org/reports/CTD_chemicals_diseases.tsv.gz'
[...]

PR #809, opened 2019-08-08
rewrite of Dataset.py to improve metadata emitted by Dipper

justaddcoffee:
To address #792, #753, #428, #545, and https://github.com/monarch-initiative/monarch-ui/issues/58

This PR will completely rewrite the Dataset.py to produce metadata that complies with [the dataset HCLS spec](https://www.w3.org/TR/2015/NOTE-hcls-dataset-20150514/)


dipper/models/Dataset.py
```diff
@@ -43,163 +181,388 @@ def __init__(
...
+
+    def _compute_triples_count(self, target_graph):
+        triples_count = len(list(target_graph.triples((None, None, None))))
+        self.graph.addTriple(self.distribution_level_turtle_curie,
+                             'void:triples',
+                             Literal(triples_count, datatype=XSD.integer))
+
+    def _compute_distinct_subjects_count(self, target_graph):
+        distinct_subjects_q = target_graph.query(
+            """SELECT(COUNT(DISTINCT ?s) as ?DistinctSubjects)
```
kshefchek: @justaddcoffee do you have a sense for how much time this adds to the ingest?  Curious how this affects some of our larger sources.  If it's slow, we should consider decoupling this from the ingest and post processing.

justaddcoffee: @kshefchek Yep, this did occur to me while I was writing this, and @TomConlin and I discussed this yesterday. It adds at least several minutes to ingests that are the size of say ZFIN. 

Tom and I decided a solution would be to add a skip_stats flag to dipper-etl.py, so that we can skip the painful sparql for ingests where time is of the essence. I did this with [this commit](https://github.com/monarch-initiative/dipper/pull/809/commits/5dd69cbf68adcc15d2d41d02b2c3235778cc4a01).

We also discussed possibly swapping out rdflib for something a little faster (redland maybe?). That library doesn't seem to be under active development though, which might cause problems down the road.

We also could postprocess, like you said, but this makes testing of the code a little more awkward, and also makes the pipeline a little less coherent IMHO. Thoughts? Glad to discuss this more if you want.

kshefchek: ah sorry, I saw that you were already addressing it.  I don't have a great answer, I still think we should just use neo4j, but if that's not an option we could use redland -- or rdflib as a last resort.  I've had good luck with the redland utils even though the project seems dead.

kshefchek: also - ntriples | cut | sort -u
if we're just counting things

justaddcoffee:
@TomConlin per our discussion, have a look at these commits

TomConlin:
General
    Note: 
    Tools I use to generate code which aspires to be PEP compliant 
    impede my ability to review code which is less compliant.

    - "Too many arguments"  happened too much to mention individually.(not much to do)
    - line too long happens to much to mention individually (please address)

    - We may not be able to stop ourselves from stuffing too many arguments 
    into every function, but we can avoid putting them all on the same line.

    - line limit 88 chars, favour leftmost continuations.
    
    - break super long constants into globals & assemble then if need be 
    - constants between ingests belong in super class
    - i.e. source logo could be a function of 'name' & constants

    - fix PEP issues introduced by this pr as found via static analysis.

    - many thanks for the renaming of legacy variables and function e.g. 'gu'
    - thanks for catching static fx w/self in panther    

Time. 
I have not done extensive timing tests and as far as I can see neither has anyone else 
I did hear mentioned "it adds a few minutes"  but I would not be shocked to learn it added a few hours as sparql is over kill for plain counting and within python is only suitable for toy scale problems.
Justin is going to add in flags to skip including the counts but that adds yet another path of not testing what goes to production. 

I will advocate that dipper lives up to is name "data ingest pipeline"and perform the counts at a later stage in the pipeline as we move towards breaking apart the monolithic catch all we are calling a pipeline.

----------------------------------------------------------------------------------
some notes on individual files


- Source
    - incorrect order of imports
    - L683 in fx declareAsOntology
    - `# TODO make sure this is synced with the Dataset class`
    - there is no better opportunity to resolve whatever this todo means.




 - HPO annotations
    ~ L177 `filedate` is declared but unused

 - MMRRC
    - L91 # TODO note: can set the data version to what is in the header
    - there is no better opportunity to resolve this todo.
    - now unused imports to remove (yea)

 - BioGrid
    - L88 `filedate` is declared but unused

 - MGISlim
    - No space allowed around keyword argument assignment L28 & L29

 - OMIA
    - are we sure ncbigene call changes are correct? looks regressive.

 - OMIMSource
    - 'self.get_files(is_dl_forced='  is 'True'  need to be 'False'

justaddcoffee:
> Time.
I have not done extensive timing tests and as far as I can see neither has anyone else
I did hear mentioned "it adds a few minutes" but I would not be shocked to learn it added a few hours as sparql is over kill for plain counting and within python is only suitable for toy scale problems.
Justin is going to add in flags to skip including the counts but that adds yet another path of not testing what goes to production.

We perhaps should just remove these SPARQL queries I added - they arguably don't really add much to the metadata and inarguably add time to the ingests.

> - fix PEP issues introduced by this pr as found via static analysis.

Can you elaborate on this item?

justaddcoffee:
> OMIA
> are we sure ncbigene call changes are correct? looks regressive.

@TomConlin can you clarify this one? I don't see a change to the ncbigene call in OMIA

TomConlin:
re your last two commits they are not part of this PR
there is a mechanism to employ when data changes and it does not involve hard coding  numbers of columns.

justaddcoffee:
> re your last two commits they are not part of this PR

Agreed, that was a little lame to hard code columns. Just trying to fix the failing ingests - it's a little hard to make sure the PR is working when the ingests from master don't run to completion.

> there is a mechanism to employ when data changes and it does not involve hard coding numbers of columns.

What is the mechanism?

TomConlin:
Looking at the dataset files from the Jenkins pipeline run 2019 Sep 16

 - we discussed a code/feature freeze for this PR last week
 - the @ prefix section does not have Monarch prefixes (please include any prefix used)
 - the `dcterms:identifier` IRI object is written as a string literal.
 - the format of the the monarch archive IRI is:
    -  `MonarchArchive/ingest_datestamp`   which is never a valid url
    - we had discussed changing to the format
    - `MonarchArchive/datestamp/rdf/ingest`  which can be a valid url if 'ttl' is changed to 'rdf'
 - we had discussed changing the directory `ttl` to `rdf` because not all output can be turtle
    - `ttl` could remain a symlink of `rdf` for backwards compatibility

 - having exact same title three times is excessive
    - please omit two or rewrite as something that justifies existence.

justaddcoffee:
> re your last two commits they are not part of this PR

Made a separate PR for these, see #825. I've also added a fix for MMRRC ingest, which was failing.

TomConlin:
the mechanism, if you have not found it yet, is the  `columns` list in the `files` dict in each ingest.  There is a fx  `Source.check_fileheader()` which provide a  shared behaviour  for what is done when _headers_ change.  

With this in place,  row levle checks are
```
 $ ag "len\(col\) != len\(row\)" dipper/sources/

dipper/sources/GWASCatalog.py
170:                if len(col) != len(row):

$ ag "len\(row\) != len\(col\)" dipper/sources/

dipper/sources/Ensembl.py
228:            if len(row) != len(col):

dipper/sources/EBIGene2Phen.py
158:        if len(row) != len(col):

dipper/sources/UDP.py
506:        # if len(row) != len(col):

dipper/sources/Coriell.py
319:                if len(row) != len(col):

dipper/sources/GeneOntology.py
269:                if len(row) != len(col):

dipper/sources/Monarch.py
129:                    if len(row) != len(col):

dipper/sources/OMIM.py
510:                if len(row) != len(col):
844:                if row and len(row) != len(col):

```

justaddcoffee:
>	•	we discussed a code/feature freeze for this PR last week

Sorry for the late edits

> 	•	the @ prefix section does not have Monarch prefixes (please include any prefix used)

Changed dataset to use MonarchArchive prefix, and to write out CURIEs (e.g. MonarchArchive:rgd) instead of IRIs. 
(commit 3693fde16dfbb7d32af4e6cbb3fe9f38c583d5d4)

>	•	the dcterms:identifier IRI object is written as a string literal.

Thanks for catching that - fixed. 
(commit 805e7e25f20a400414ba51e871919097fb82e9b9)

>	•	the format of the the monarch archive IRI is:
>	◦	MonarchArchive/ingest_datestamp which is never a valid url
>	◦	we had discussed changing to the format
>	◦	MonarchArchive/datestamp/rdf/ingest which can be a valid url if 'ttl' is changed to 'rdf'

Yep, I like the new URL pattern MonarchArchive/datestamp/rdf/ingest

The downloadURL I’m writing in this PR conforms to what we discussed I think. It’s pointing to:
http://archive.monarchinitiative.org/[version]/rdf/[source].[distribution file type]
E.g. Here’s the downloadURL triple I’m writing for RGD:
dcterms:downloadURL <https://archive.monarchinitiative.org/201908/rdf/rgd.ttl> ;

The summary, version, and distribution level IRIs I'm writing are identifiers that I think are coherent, but obviously are not URLs currently (i.e. they don’t point to a file or directory that exists).
Summary level IRI:
<https://archive.monarchinitiative.org/[source]>
Version level IRI:
<https://archive.monarchinitiative.org/[source]_[version]>
Distribution level IRI:
<https://archive.monarchinitiative.org/[source]_[version].[distribution file type]>

For example for RGD these are:
<https://archive.monarchinitiative.org/rgd>
<https://archive.monarchinitiative.org/rgd_201908>
<https://archive.monarchinitiative.org/rgd_201908.ttl>

Could you let me know what you’d prefer this for the IRIs? These conform more to what you are talking about, but only the distribution level IRI points to anything real.
<https://archive.monarchinitiative.org/rgd> # don’t think there is a way to make this point to a file or directory that actually exists
<https://archive.monarchinitiative.org/201908/rdf/rgd> # also not pointing to anything real, but pointing to a directory that exists
<https://archive.monarchinitiative.org/201908/rdf/rgd.ttl> # pointing to something downloadable
I think these are a little less meaningful than the present IRIs I’m writing in this PR, but glad to change them if you like.

>	•	we had discussed changing the directory ttl to rdf because not all output can be turtle
>	◦	ttl could remain a symlink of rdf for backwards compatibility

Yes, agreed - it’s good if TTLs and NTs live in the same directory, and also good if we keep this backward compatible.

>	•	having exact same title three times is excessive
>	◦	please omit two or rewrite as something that justifies existence.

I’ve appended the version number to the version level title, and the file type to the distribution level title. I don’t think we can omit two - all three titles must be present according to the spec. 
(commit 13a79c76975435cf348a5016eecca1e2339a0121)

justaddcoffee:
For my own info, two other tickets that are possibly addressed by this PR:
https://github.com/monarch-initiative/dipper/issues/428
https://github.com/monarch-initiative/dipper/issues/545

TomConlin:
> Okay, thanks. This makes sense, although there was/is no column info in AnimalQTLdb.

Thanks, there  was/is no column info in any ingest till very recently which is why we add it in when we have the file open.

> E.g. Here’s the downloadURL triple I’m writing for RGD:
> dcterms:downloadURL https://archive.monarchinitiative.org/201908/rdf/rgd.ttl ;

That URL now resolves!  (as will any  of the .ttl or .nt)

----------------------------------------------
More helpful would be better, 
we do not have version and summary for specific ingests, 
because they are not what we are producing.

Ingest rdf are a necessary subset of what we do produce 
which is an episodic dipper release.

I expect the only legitimate summary and version data for the ingest 
would have to be from the source, about the source, and although some ingests 
will have more than enough metadata to support these dataset object others will not. 
The only certainty is all will be different and out of our control. 
(read more perpetual maintenance)
  
For this I would argue our Version and Summary metadata URLs should be to 
Dipper release products (despite being repeated in each ingest)

Distribution level URL 
    resolving to the download file I think is great.

Version  level URL 
    https://archive.monarchinitiative.org/[version]
    Could have a static or generated index.html to provide helpful wordage about our 
    insufficiently frequent release process.
    

Summary level URL
    https://archive.monarchinitiative.org
    *Should* have a static index.html to provide helpful wordage 
    about the datasets provided.

justaddcoffee:
> More helpful would be better,
> we do not have version and summary for specific ingests,
> because they are not what we are producing.

I guess I disagree with this. Each ingest is run separately and is emitted as a separate TTL|NT file, so I think it makes sense to have separate version and summary IRIs for each separate ingest of each source. This distinction seems useful/appropriate. For example, if we use the same version IRI for all ingests in a given Dipper data release version (say, https://archive.monarchinitiative.org/201909/), each of the 36 or so TTL files we emit in each version will claim to be this generic version of the Monarch dataset, but they'll all be different.

Also, if we conflate the version and summary IRIs for all ingests, we remove the ability to refer to the individual ingests separately further down the stack. It seems possible that this ability will be useful downstream when we use this metadata for provenance.

@cmungall thoughts?

TomConlin:
We only loose the ability to refer to the individual ingests separately if we choose not to use the distribution level URL. 
Appending the ingest-name to the intermediate level URL for Summary and Version is fine.  
maybe they can be made useful some day

cmungall:
Sorry, will try and give this some thought next week - is this a blocker?

justaddcoffee:
> Sorry, will try and give this some thought next week - is this a blocker?

No worries - not really a blocker, more of a request for a philosophical second opinion.

TomConlin:
+1

PR #815, opened 2019-08-22
Comment out gene to disease parser for ctd and kegg

### LINKED ISSUE #814: Remove gene to disease ingest from CTD and KEGG
Remove gene to disease data from CTD and KEGG ingests

kshefchek:
Per @cmungall 's request - removing the gene to disease parsers from kegg and CTD.  Closes #814


kshefchek:
I got a little overzealous, you're right we need these for the clique merge, added everything back but process_kegg_disease2gene

PR #816, opened 2019-08-22
Run omim sources serially

kshefchek:
Some of our sources utilize the OMIM dataset to determine if an OMIM id is a disease or a gene.  Running these all in parallel causes us to query OMIM at too frequently.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -47,7 +47,6 @@ pipeline {
 
                         # Clean up previous runs
                         sudo rm -rf ./out/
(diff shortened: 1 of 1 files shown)


TomConlin:
you have all the ones I know about

```
ag -l OMIMSource
dipper/sources/GeneReviews.py
dipper/sources/NCBIGene.py
dipper/sources/HGNC.py
dipper/sources/OMIA.py
dipper/sources/OMIM.py
dipper/sources/OMIMSource.py
dipper/sources/KEGG.py
```

PR #729, opened 2019-04-25
updates to ClinVar to support RCVs with multiple conditions

kshefchek:
Adjusts the ClinVar parser to accommodate the latest xml schema, specifically:
- RCVs can relate a variant/genotype to 1 or more conditions
- RCVs associate either a MeasureSet (Variant) or GenotypeSet to the disease(s)
- A MeasureSet has 1 or more Measures (allele), a GenotypeSet has 1 or more MeasureSets

Makes the following assumptions/changes
- A genotype/variant is associated to every disease (RCV and SVC)
- Distinct chromosomes = mosaic genotype, we need to add a mosaic genotype term in GENO.  NCIT contains a term but it is a subclass of organism, whereas we want a sequence feature.
- Changed xrefs to dbSNP from dbXref to owl:sameAs

Things we are not parsing
- Variants and diseases marked as co-occurring.  For co-occurring diseases a variant is associated to both diseases
- Haploinsuffiency and triploinsufficiency data, this could be more easily parsed from another source (ClinGen)
- Phase unknown is typed as a haplotype

Tests:
CompoundHet, same gene
https://www.ncbi.nlm.nih.gov/clinvar/RCV000162061/

Co-occurring diseases, co-occurring variants
https://www.ncbi.nlm.nih.gov/clinvar/RCV000416376/

Multiple similar conditions
https://www.ncbi.nlm.nih.gov/clinvar/RCV000763295/

Variant overlaps two genes, and only affects one (exon of one, intron of other)
https://www.ncbi.nlm.nih.gov/clinvar/RCV000175394/

"Typical" RCV, multiple SCVs, BRCA1
https://www.ncbi.nlm.nih.gov/clinvar/RCV000112698/

Haplotype, >1 genes, two copy number gain variants, Bardet-Biedl syndrome (BBS)
https://www.ncbi.nlm.nih.gov/clinvar/RCV000498447/

NM_000090.3(COL3A1):c.[=/3426_3452del27];[=/3440_3466del27] AND Ehlers-Danlos syndrome, type 4
Mosaic for two separate 27bp deletions in exon 48 in two cell strains of equal proportion
https://www.ncbi.nlm.nih.gov/clinvar/RCV000087646/

Output turtle for tests:
https://github.com/monarch-initiative/dipper/tree/clinvar-RCV000416376/tests/resources/clinvar/output

Output graphviz for tests:
https://github.com/monarch-initiative/dipper/tree/clinvar-RCV000416376/docs/img

[...]

PR #819, opened 2019-08-26
Make clinvar variants and dbsnp ids xrefs instead of eqs

kshefchek:
In https://github.com/monarch-initiative/dipper/pull/729 I made ClinVar variants and their xref dbSNP ids equivalent, whereas previously they were xrefs.  This has caused a number of clique merge bugs in the latest load, for example:
https://www.ncbi.nlm.nih.gov/clinvar/variation/31915/
https://www.ncbi.nlm.nih.gov/clinvar/variation/21303/

### DIFF
--- a/dipper/sources/ClinVarXML_alpha.py
+++ b/dipper/sources/ClinVarXML_alpha.py
@@ -361,12 +361,16 @@ def allele_to_triples(allele, triples) -> None:
     if allele.label is not None:
         write_spo(allele.id, 'rdfs:label', allele.label, triples)
 
-    # <ClinVarVariant:rcv_variant_id><OWL:sameAs><dbSNP:rs>
+    # <ClinVarVariant:rcv_variant_id><OWL:hasDbXref><dbSNP:rs>
+    #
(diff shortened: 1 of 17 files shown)


dipper/sources/ClinVarXML_alpha.py
```diff
@@ -361,12 +361,16 @@ def allele_to_triples(allele, triples) -> None:
...
+    # Note that making clinvar variants and dbSNPs equivalent
+    # causes clique merge bugs, so best to leave them as xrefs
+    # Example: https://www.ncbi.nlm.nih.gov/clinvar/variation/31915/
+    # https://www.ncbi.nlm.nih.gov/clinvar/variation/21303/
     for dbsnp_id in allele.dbsnps:
-        # sameAs or hasdbxref?
         write_spo(
             allele.id,
-            'owl:sameAs',
+            'oboInOwl:hasDbXref',
```
TomConlin: `self.globaltt['database_cross_reference']`

@kshefchek  always use the global mapping

kshefchek:
i'm going to open a new pr for this

PR #828, opened 2019-09-25
Modify dataset IRIs

justaddcoffee:
This PR alters the dataset summary, version, and distribution IRIs from:
```
https://archive.monarchinitiative.org/[source]
https://archive.monarchinitiative.org/[source]_[version]
https://archive.monarchinitiative.org/[source]_[version].[distribution file type]
```
To what Tom suggests:
```
https://archive.monarchinitiative.org/#[source]
https://archive.monarchinitiative.org/[version]/#[source]
https://archive.monarchinitiative.org/[version]/rdf/[source].[distribution type]
```
For example for RGD these IRIs will now be: 

```
https://archive.monarchinitiative.org/#rgd
https://archive.monarchinitiative.org/201908/#rgd
https://archive.monarchinitiative.org/201908/rdf/rgd.ttl
```

n.b. Dipper doesn't consider the CURIE versions of these IRIs (e.g. `MonarchArchive:/#rgd`) to be valid CURIEs, but will write them out anyway if told to do so

### DIFF
--- a/dipper/models/Model.py
+++ b/dipper/models/Model.py
@@ -218,7 +218,10 @@ class Model():
             ontology_id, self.globaltt['type'], self.globaltt['ontology'])
 
     def addOWLVersionIRI(self, ontology_id, version_iri):
-        self.graph.addTriple(ontology_id, self.globaltt['version_iri'], version_iri)
+        self.graph.addTriple(ontology_id,
+                             self.globaltt['version_iri'],
+                             version_iri,
+                             object_is_literal=False)
 
     def addOWLVersionInfo(self, ontology_id, version_info):
         self.graph.addTriple(
--- a/dipper/models/Dataset.py
+++ b/dipper/models/Dataset.py
@@ -164,7 +164,7 @@ class Dataset:
 
     def __init__(
             self,
-            identifier,       # name? should be Archive url via Source
+            identifier,
             data_release_version,
             ingest_name,
             ingest_title,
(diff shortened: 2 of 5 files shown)


dipper/models/Model.py
```diff
@@ -218,7 +218,10 @@ def addOntologyDeclaration(self, ontology_id):
             ontology_id, self.globaltt['type'], self.globaltt['ontology'])
 
     def addOWLVersionIRI(self, ontology_id, version_iri):
-        self.graph.addTriple(ontology_id, self.globaltt['version_iri'], version_iri)
+        self.graph.addTriple(ontology_id,
+                             self.globaltt['version_iri'],
+                             version_iri,
+                             object_is_literal=False)
```
justaddcoffee: @TomConlin interested in your opinion here. I'm having to add this to force the dataset version IRI (e.g. `MonarchArchive:/201908/#OMIM`) to be considered an IRI and not a literal. This change means that any ingest that sends a OWL version IRI that isn't a curie will fail with an `AssertionError` when Dipper goes to actually add the triple. That is okay I think, I'm just curious where you suggest we draw the line about when to make an ingest fail versus emit a warning and recover.

TomConlin: I have done the same but within various helper functions that are intended to only emit non literal objects. What do you mean it must be a curie?  Seems sending a full iri should work as well as a curie at not being a literal.

justaddcoffee: > What do you mean it must be a curie? Seems sending a full iri should work as well as a curie at not being a literal.

Yep, should've said CURIE or IRI

> I would make it fail if version_iri could not be a proper IRI or a curie.

Okay, good, thanks. That's the behavior here

TomConlin:
can re open if it is decided we should do something else

PR #829, opened 2019-09-26
pylint Makefile tests

TomConlin:
Firsts steps towards  shared static checking by default.

not enabled/enforced to begin with (tests would fail)

At minimum I would like to see `make lint_error` always pass
and move towards `make lint_warn` producing minimal output.

 - `.pylintrc` may be edited to tune for our collective tolerance for cruft
 - specialized Malkfile rules can be added to suit particular cases.


justaddcoffee:
@TomConlin made a slight change to `./scripts/within_pip_env.py`

Very much like/appreciate a set of pylint checks we can do in `Makefile` format

dipper/curie_map.yaml
```diff
@@ -193,6 +193,7 @@
 'dictyBase': 'http://dictybase.org/gene/'                                       # Dictyostelium discoideum database (social amoebae)
 'EcoGene': 'http://ecogene.org/gene/'                                           # Escherichia coli K-1
 'ENSEMBL': 'http://ensembl.org/id/'                                             # joint EMBL, EBI, Sanger Institute central resource
+'EnsemblGenome': 'http://www.ensemblgenomes.org/id/'                            # The Ensembl genome annotation system
```
justaddcoffee: Why the new curie prefix? (just curious)

TomConlin:
It is disturbing that something so purposefully basic fails on a platform being used for development.

dipper/curie_map.yaml
TomConlin: they were there before I found them, just written out as literals.

PR #831, opened 2019-09-27
(Don't merge) Capture review status and "star" info from ClinVar

justaddcoffee:
First crack at capturing review status and scores, with the goal of allowing ClinVar records to be filtered by "star" rating, similar to how Exomiser apparently does this filtering.

I don't have SEPIO terms for the review statuses yet (waiting on this ticket https://github.com/monarch-initiative/SEPIO-ontology/issues/30), so I've made some up:
```
"no assertion criteria provided": "SEPIO:9999990",
"no assertion provided": "SEPIO:9999991",
"criteria provided, single submitter": "SEPIO:9999992",
"criteria provided, conflicting interpretations": "SEPIO:9999993",
"criteria provided, multiple submitters, no conflicts": "SEPIO:9999994",
"reviewed by expert panel": "SEPIO:9999995",
"practice guideline": "SEPIO:9999996",
```
Obviously these will need to be replaced.

Also, tests are failing because I haven't updated ClinVar's test graphs to include the new triples I'm writing.


kshefchek:
looks good!  I think we want the RCV (aggregate variant to disease) review status as opposed to the SCV (single submission for variant to disease), and have this attached to the oban association.

justaddcoffee:
> looks good! I think we want the RCV (aggregate variant to disease) review status as opposed to the SCV (single submission for variant to disease), and have this attached to the oban association.

@kshefchek have a look if you get a chance, I think this last commit is what we want

dipper/sources/ClinVarXML_alpha.py
```diff
@@ -476,6 +477,68 @@ def record_to_triples(rcv: ClinVarRecord, triples: List, g2p_map: Dict) -> None:
...
+        "reviewed by expert panel": '3',
+        "practice guideline": '4',
+    }
+    for status, score in status_and_scores.items():
+        triples.append(
+            make_spo(
+                GLOBALTT[status],
+                GLOBALTT['has specified numeric value'],
+                score))
+    return triples
```
kshefchek: does this end up in the output?

justaddcoffee: > does this end up in the output?

Yep, should be output via [this line](https://github.com/justaddcoffee/dipper/blob/c40c1fc3c88b2aa9284b25f7d3f4f4fa0c8038f4/dipper/sources/ClinVarXML_alpha.py#L644)

PR #836, opened 2019-10-08
fix broken import in OMIM.py

justaddcoffee:
USER_AGENT doesn't exist in OMIM.py, I think this is supposed to be imported from OMIMSource


TomConlin:
+1

PR #839, opened 2019-10-11
Several fixes for Mac:

justaddcoffee:
- made test for sortedness of translation table case insensitive, which I think is what we want
- removed gnu specific -f flag from time command in lint* tests
- removed @ from test_source_metadata and test_wormbase, b/c I assume we don't really want to suppress output from these

Also, GLOBAL_TERMS.yaml apparently wasn't sorted according to
sort -k2,2 -k3,3n -t ':' --stable
because of these comment lines:
```
# ??? PATO_0000470: GENO:0000876
# Place holder predicates: predicates
# found in sepio_developer.owl:  SEPIO:0000019
# is not found in sepio:  SEPIO:0000111 --
```
I've removed them. A general comment is that enforcing this sort
is going to make comments in GLOBAL_TERMS.yaml a bit awkward


translationtable/GLOBAL_TERMS.yaml
```diff
@@ -554,8 +552,6 @@
 "is causal somatic mutation in": "RO:0004014",
 "is causal susceptibility factor for": "RO:0004015",
 "is causal germline mutation partially giving rise to": "RO:0004016",
-# found in sepio_developer.owl:  SEPIO:0000019
```
kshefchek: let's leave these comments in

translationtable/GLOBAL_TERMS.yaml
```diff
@@ -179,7 +179,6 @@
 "organization": "foaf:organization",
 "page": "foaf:page",
 "person": "foaf:Person",
-# ??? PATO_0000470: GENO:0000876
```
kshefchek: Leave in and expand, @TomConlin do you recall what this comment means

kshefchek:
I don't understand the nuance between gnu sort on debian derivatives vs sort on macOS, it seems some comments are okay but others are not, @justaddcoffee do you know why these four comment lines resulted in the check failing?

translationtable/GLOBAL_TERMS.yaml
TomConlin: the geno  seems to be "obsolete_genetic dosage" 
the pato is  "increased amount"
I imagine the geno had been hardcoded in the source somewhere and was moved 
the tt  where it waits to be resolved  the pato may have been a SWAG at a replacement but remained insufficient to act on (requires bio input?).

There may be too much of a disconnect between terms and labels in rapidly evolving ontology to be sure the current obsolete label is relevant to the the previously hardcoded term.

Issue #823, opened 2019-09-02
Rename ClinVar module and update docs


cmungall:
Trying to figure out clinvar ingest from dipper docs.

It points to:

https://dipper.readthedocs.io/en/latest/dipper.sources.ClinVarXML_alpha.html

Generally not a good idea to embed release state in the file/module name.

Docs say:

_clinvarxml_alpha First pass at converting ClinVar XML into RDF triples to be ingested by SciGraph. These triples conform to the core of the SEPIO Evidence & Provenance model 2016 Apr_

this doesn't look so good if we are pointing people at our docs to figure out our modeling

TomConlin:
>Generally not a good idea to embed release state in the file/module name.


https://github.com/monarch-initiative/dipper/pull/819/files#r317832863

PR #840, opened 2019-10-17
Rename clinvar source, make dbsnp dbxref

kshefchek:
This is similar to https://github.com/monarch-initiative/dipper/pull/819

except I have renamed the clinvar parser and am using the global translation table for database cross reference.  Notes from before:

In #729 I made ClinVar variants and their xref dbSNP ids equivalent, whereas previously they were xrefs. This has caused a number of clique merge bugs in the latest load, for example:
https://www.ncbi.nlm.nih.gov/clinvar/variation/31915/
https://www.ncbi.nlm.nih.gov/clinvar/variation/21303/

Fixes https://github.com/monarch-initiative/dipper/issues/823


tests/resources/clinvar/expected/RCV000175394.ttl
```diff
@@ -4,7 +4,7 @@
     OBAN:association_has_object <http://omim.org/entry/609016> ;
     OBAN:association_has_predicate OBO:GENO_0000841 ;
     OBAN:association_has_subject <http://www.ncbi.nlm.nih.gov/clinvar/variation/194917> ;
-    oboInOwl:hasdbxref ClinVar:RCV000175394 .
+   oboInOwl:hasDbXref ClinVar:RCV000175394 .
```
TomConlin: leading space?

TomConlin: +1

kshefchek: updated

PR #853, opened 2019-10-28
NCBIGene fix omim dbxref

TomConlin:
Curie v.s. omim number mismatch
adds 16,076 omim dbxref

### DIFF
--- a/dipper/sources/NCBIGene.py
+++ b/dipper/sources/NCBIGene.py
@@ -423,10 +423,11 @@ class NCBIGene(OMIMSource):
                                 model.addXref(gene_id, dbxref_curie)
                                 omim_num = omim  # last wins
 
-                    elif omim_num in self.omim_type:  # and
-                            # self.omim_type[omim_num] == self.globaltt['gene']:
+                    elif omim_num in self.omim_type and\
+                            self.omim_type[omim_num] == self.globaltt['gene']:
                         model.addXref(gene_id, dbxref_curie)
-
+                    else:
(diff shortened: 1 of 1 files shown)


dipper/sources/NCBIGene.py
```diff
@@ -388,38 +388,47 @@ def _add_gene_equivalencies(self, xrefs, gene_id, taxon):
...
+                    omim_num = dbxref_curie[5:]
+                    if omim_num in self.omim_replaced:
+                        repl = self.omim_replaced[omim_num]
                         for omim in repl:
                             if omim in self.omim_type and \
                                     self.omim_type[omim] == self.globaltt['gene']:
-                                dbxref_curie = omim
-                    if dbxref_curie in self.omim_type and \
-                            self.omim_type[dbxref_curie] != self.globaltt['gene']:
-                        continue
```
kshefchek: We need to avoid making ncbi genes and non OMIM genes equivalent, which is why the continue is here, without it they are made equivalent in the try block below

TomConlin: I see,  will else continue

TomConlin:
dropped 1,400  ncbigene <-> omim-notgene

```
NCBIGene -> OMIM [label="oboInOwl:hasDbXref (16078)", color="blue"];
NCBIGene -> OMIM [label="owl:equivalentClass (-67)", color="black"];
NCBIGene -> OMIM [label="owl:sameAs (-1329)", color="black"];
```

PR #854, opened 2019-10-28
Fix for GO ingest

justaddcoffee:
In `zfin.gaf.gz`, very large entry in field 7 evidently breaks `csv`'s ability to parse with default field size limit. Increasing this fixes it


TomConlin:
it looks benign,  have you run it? 
I am currently about an hour and a half into testing my PR and am not keen on testing this one too.

justaddcoffee:
> it looks benign, have you run it?
> I am currently about an hour and a half into retesting my PR and am not keen on testing this one too.

Yep, I've run it, runs to completion on my machine

TomConlin:
awesome +1

PR #855, opened 2019-10-28
Fix monarch owl build step in Dipper pipeline

justaddcoffee:
Updated broken link to owltools to point to a working link for (and known version) of owltools. This link is broken for the time being because of power outages here at LBL. @kltm advised that the link we were using, even when it is working, is probably not a good choice, as it is likely to be volatile. 

(We could alternatively point to http://current.geneontology.org/bin/owltools, but we'd obviously be linking to an unknown version of owltools.)

Also, we possibly could/should be using robot instead of owltools here


kltm:
Generally, I'd assume that owltools should be avoided if there are other tools like ROBOT, as @justaddcoffee mentions above.
If owltools cannot be avoided, we'd like to have a system where a version could be build or downloaded from maven central, etc. We do not, currently, have that and that's Bad.
If owltools is required, the build.berkeleybop.org URL is known to be volatile  and may be changed in the future anyways; as an alternate hack, the URL as given above may also be used as needed as a workaround.

kshefchek:
@kltm a robot pipeline to replace this is in the works

PR #858, opened 2019-11-01
Update ontology term to one that exists 

TomConlin:
`assertion_confidence_level` exists in OLS `confidence_score` does not
then fix tests

### DIFF
--- a/dipper/sources/ClinVar.py
+++ b/dipper/sources/ClinVar.py
@@ -973,10 +973,10 @@ def parse():
                     # to this association to allow filtering of RCV by review status
                     if rcv_review is not None:
                         write_spo(
-                                  monarch_assoc,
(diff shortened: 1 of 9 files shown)


kshefchek:
confidence_score is a datatype property and assertion_confidence_level is an object property, and I also have SEPIO:0000168 configured in scigraph and solr, prefer if we kept this for now

TomConlin:
@mbrush  having updates in OLS/Ontobee is out best path forward.

kshefchek:
It looks like SEPIO_0000168 has been in the ontology dating back to 2017, Matt do you know why it's not in OLS?

matentzn:
Seems like SEPIO is still on the 2017 version, and 168 was introduced later. @mbrush and I will make a new release soon I hope, and then this issue will be fixed.

PR #862, opened 2019-11-07
CTD fix & refactor

TomConlin:
Was only emitting metadata.
refactor to current conventions which should help ingest to stop when it is broken
not convinced file time comparison is avoiding unnecessary download

(leftover minor edits from our keeping unsupported term in clinvar)


kshefchek:
+1

PR #863, opened 2019-11-11
UCSCBands (& Monochrom)

TomConlin:
Make UCSCBands produce output again in the course of refactoring.
As these are not "used"  in scigraph and in the case of Monochrom 
not really "dipper data" a-box stuff so they tend to be overlooked.
 - update existing assembles to latest build
 - add cat & dog  on account of because I wanted to
 - set up to add eight-ish other species well represented in the rest of dipper 
 - refactoring UCSC to make more readable better breadcrumbs back to source
 - minimal change to Monochrom for it to run.
 
There are outstanding ontology issues, but the current state is not worse than it was.


dipper/sources/UCSCBands.py
```diff
@@ -151,73 +211,76 @@ def __init__(self,
...
-    def _get_chrbands(self, limit, taxon, genome_id):
+    def _get_chrbands(self, limit, src_key, genome_id):
         """
         :param limit:
         :return:
 
         """
-
+        tax_num = src_key
+        species_name = self.globaltcid[tax_num]  # for logging
```
justaddcoffee: Getting a key error here, this I think should be `self.globaltcid["NCBITaxon:" + tax_num]`

TomConlin: thanks @justaddcoffee  good catch

PR #867, opened 2019-11-20
Fix for OMIM ingest

justaddcoffee:
Column name changed from 'Mim Number' to 'MIM Number'

Makefile is broken on Mac, no idea if this passes unit tests

### DIFF
--- a/dipper/sources/OMIMSource.py
+++ b/dipper/sources/OMIMSource.py
@@ -32,7 +32,7 @@ class OMIMSource(Source):
             'clean': OMIMURL,
             'columns': [  # expected
                 'Prefix',
-                'Mim Number',
+                'MIM Number',
                 'Preferred Title; symbol',
                 'Alternative Title(s); symbol(s)',
                 'Included Title(s); symbols',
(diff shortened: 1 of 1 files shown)


TomConlin:
Travis (which you could enable on you personal fork as well) says it passes.

justaddcoffee:
Duly noted, but It would be nice to be able to tell without pushing to GH

kshefchek:
Docker is always an option, albeit with more steps than just hitting make.
    
    cd /path/to/git/dipper
    docker run \
        -it \
        -v $PWD:/dipper \
        --entrypoint "/bin/bash" \
        python:3.6-buster

    cd dipper && python3 -m venv venv
    source venv/bin/activate
    pip install -r requirements.txt
    pip install -r requirements/all-sources.txt
    make

Theres also coreutils on mac.  Still surprising to me that we can't figure this out, but the number of stackoverflows related to this makes me doubt that we'll get 100% consistency across OSs

justaddcoffee:
I can just upgrade bash and sort on my mac I think, not a huge deal, but possibly should document somewhere that MacOS users will need to do this to get Dipper to work

PR #876, opened 2019-12-12
Change ensembl URL and emit error message when non-200 is received

justaddcoffee:
Two changes: 
- switched to 'www.ensembl.org'  from 'uswest.ensembl.org' - the latter seems to have not worked for the last few days
- emit an error message when querying from `/biomart/martservice` fails (this was silently failing previously)


TomConlin:
> switched to 'www.ensembl.org' from 'uswest.ensembl.org' - the latter seems to have not worked for the last few days

The only observable difference I notice here is the deletion of the comment that provides the original/alternative url.   see:  
https://github.com/monarch-initiative/dipper/pull/876/files#diff-15c481a6eeaf8dcc8ff228ecac079919L13

adding checks on return http status codes is an unmitigated good.  +1

Unless `www.ensembl.org` is permanently retired. 
there no benefit from removing the comment I could use 
in a trouble shooting situation.

justaddcoffee:
> The only observable difference I notice here is the deletion of the comment that provides the original/alternative url

Sorry, neglected to commit what I was actually using locally. Actually updated URL to `www.ensembl.org` with last commit

TomConlin:
perfect. thanks

justaddcoffee:
Thanks Tom!

PR #887, opened 2020-01-08
Set created on and data release version in metadata dynamically

justaddcoffee:
Two minor changes: 
- set `dcterms:created` using actual date instead of DATA_RELEASE_VERSION
- set DATA_RELEASE_VERSION in Jenkins using call to `date`


TomConlin:
Looks good to me

PR #888, opened 2020-01-09
Fix metadata (dct:source and pav:retrieved_on) for files retrieved from DipperCache

justaddcoffee:
For source files retrieved from DipperCache, currently we are missing metadata such as: 

```
<http://purl.obolibrary.org/obo/zp//id_map_zfin.tsv> pav:retrievedOn "2019-12-02T09:38:46"^^xsd:dateTime .
```
and: 

```
<https://archive.monarchinitiative.org/20200109/#zfin> 
    dcterms:source
        <https://archive.monarchinitiative.org/DipperCache/zfin/id_map_zfin.tsv>,
```

### DIFF
--- a/dipper/sources/Source.py
+++ b/dipper/sources/Source.py
@@ -368,13 +368,17 @@ class Source:
             headers = None
             filesource = files[src_key]
             # attempt to fetch from a web cache
-            if self.fetch_from_url(
-                    '/'.join((self.DIPPERCACHE, self.name, filesource['file'])),
-                    '/'.join((self.rawdir, filesource['file'])),
-                    is_dl_forced):
+            remote_file = '/'.join((self.DIPPERCACHE, self.name, filesource['file']))
+            local_file = '/'.join((self.rawdir, filesource['file']))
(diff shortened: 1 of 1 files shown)


TomConlin:
I am wonder about overlap between the existing
[check_if_remote_is_newer()](https://github.com/monarch-initiative/dipper/blob/master/dipper/sources/Source.py#L274)
and your  new `get_remote_last_modified()` 
seems if a date exists, we have already asked for it and may just need to retain it

justaddcoffee:
> I am wonder about overlap between the existing
> [check_if_remote_is_newer()](https://github.com/monarch-initiative/dipper/blob/master/dipper/sources/Source.py#L274)
> and your new `get_remote_last_modified()`
> seems if a date exists, we have already asked for it and may just need to retain it

Have a look, removed get_remote_last_modified() and saved these timestamps within check_if_remote_is_newer(), then used them to write the retrievedOn triple

TomConlin:
seems reasonable to me +1

PR #899, opened 2020-01-21
Add entry in curie_map.yaml for ComplexPortal data ingested in GO

justaddcoffee:


TomConlin:
Looks good, 
note:  there is currently an upper bound of 325 possible instances

PR #905, opened 2020-02-10
Fix retrieved_on subject for items retrieved from DipperCache

justaddcoffee:
I previously was using URL for remote file, not DipperCache URL as the subject for this triple.


TomConlin:
JR> # Here we assume that the timestamp of the cached file on
JR> # DipperCache is the date it was retrieved from the upstream source

TC> # Here the timestamp on the file in DipperCache is a best effort representation
TC> # of the earliest time the file could have been retrieved from source.

if the updated assumption still holds then the pr is good to go

Issue #910, opened 2020-02-20
Flybase ingest incorrectly parses allele to phenotype lines without an anatomical or GO term


kshefchek:
For example given the line:

    FBal0038994	dendritic spine & lobular plate tangential neuron, with @FBal0220362:Scer\GAL4<up>DB331</up>@	derived_pheno_manifest	FBrf0206747	Characterization of dendritic spines in the Drosophila central nervous system.	19160442

We generate the incorrect triple

    FlyBase:FBal0038994 RO:0002200 OBO:FBal0220362PHENOTYPE

This is from:
https://github.com/monarch-initiative/dipper/blob/299af4e32/dipper/sources/FlyBase.py#L301

We should get rid of this entire chunk of code given #909

Issue #770, opened 2019-06-20
Question: how to ingest and model "phenotype manifest in" data from FlyBase


kshefchek:
1. How do we handle "phenotype manifest in" data?
- Link them to the anatomical entity via 'located in' or 'expressed in' and/or;
- Link them to UPheno classes with 'has phenotype', eg https://monarchinitiative.org/phenotype/FBbt:00000004PHENOTYPE or;
- Ignore all together

kshefchek:
See also https://github.com/monarch-initiative/dipper/issues/248

dosumis:
All should be interpreted as phenotypes following the [abnormalAnatomicalEntity pattern](
https://github.com/obophenotype/upheno/blob/master/src/patterns/dosdp-patterns/abnormalAnatomicalEntity.yaml) - I guess this is option 2?

In a small number of cases we may be able to use qualifiers (supernumarary, ectopic) to infer the type of abnormality.  Some have stage too (FBdv), but that should probably be folded in  to annotation  if possible.

Two options for generation.

1. Shadow all of FBbt - generating terms using the above pattern.
2. Shadow only terms used in phenotype_manifest_in.

Think I prefer the latter.   @matentzn  - this would be like a much simpler, stripped down version of ZP/XP.   Not sure worth releasing as a separate artefact for others to use in annotation though.  

TBD: ID managment.

matentzn:
This will require a bit more thought. Can we go with 2, but leave the issue open until David and I found a more resilient solution?

PR #911, opened 2020-02-20
FlyBase: skip derived_pheno_manifest in parsing allele phenotype association file

kshefchek:
Fixes https://github.com/monarch-initiative/dipper/issues/910

See also https://github.com/monarch-initiative/dipper/issues/770


kshefchek:
Oddly, travis is not showing a failure when our tests are failing: https://travis-ci.org/monarch-initiative/dipper/builds/653194301, this PR definitely breaks the flybase tests

TomConlin:
it is because the python virtual env check eats the error

https://github.com/monarch-initiative/dipper/blob/master/Makefile#L13

we can likely find a way to both have an informative message and propagate the error

PR #912, opened 2020-02-21
Makefile error propagation and misc lint  & minor refactor

TomConlin:
My checking if within a python virtual env ate the errors travis needed to break a build on unit test failure.  errors were displayed but did not stop the world.  

move more owl file fetching from ad-hoc  locations to obo.

change from maintaining  a set 
to appending to a list, then making a set once when done.

rest is comments and whitespace

moved a script never used here to transdipperator 

reactome & worm fail travis tests (but not local tests or production runs) 
because they relied on files they did not fetch existing.


TomConlin:
Yea!  it fails hard again!

```
Makefile:51: recipe for target 'test_impc' failed

make: *** [test_impc] Error 1

The command "make" exited with 2.

Done. Your build exited with 1.

```

kshefchek:
The test passes, try:

python -m unittest tests/test_impc.py

but fails with

    make test

TomConlin:
Had omitted a pair of parens please try agaun

PR #913, opened 2020-02-27
OBO url https -> http 

justaddcoffee:
Changed 
        `https://purl.obolibrary.org/obo`
to 
        `http://purl.obolibrary.org/obo`
The former apparently doesn't resolve, causing lots of ingests to fail


TomConlin:
In this day and age the bug is there not here

PR #914, opened 2020-02-27
Fix key error in Reactome ingest

justaddcoffee:


TomConlin:
would rather see a preceding assignment

```
src_key = 'gaf-eco-mapping'
```
the pattern now exists in 14 ingests.
thought being  with with over 30 instances, these sections are candidates for factoring out

TomConlin:
+1

PR #916, opened 2020-03-02
panther - type genes as gene instead of owl:Class

kshefchek:
Replaces triples that define genes as an owl:Class instead with rdf:type globaltt['gene']


TomConlin:
looks good,  I think most of this typing should be redundant

kshefchek:
check out https://archive.monarchinitiative.org/201910/qc/category-pivot.tsv (a little outdated), categorizes are the most informative categories, so 'Node' means that we're not categorizing some of these as genes.

PR #917, opened 2020-03-16
UDP  & misc

TomConlin:
UDP (which I cannot run) got explicit MONARCH: curie prefix. Relying on default base prefix  results in a colon where we expect an underscore when processed with standard tools. (remember to fix tests)

Trying a run without duplicating metadata in the main graph, should eliminate whole classes of edge cases in post processing. 

Far fewer hardcoded curiemap  exceptions (for graphviz dot) theses days.

recently added try/except referenced non existent  .read() function 

spelling in comments type corrections


TomConlin:
that is odd, I do not get a make fail on dataset unit test locally.

PR #920, opened 2020-03-27
Global terms label updates mostly

TomConlin:
remove stray char in jenkins file
bringing global term labels closer to published labels 
back off on tagging git per release
don't move compressed archive in alongside  the same uncompressed files
use builtin line counter in animalqtl

update several term labels


kshefchek:
+1

PR #924, opened 2020-04-14
Revert Gallus gallus URL from Dipper Cache to NCBI URL

justaddcoffee:
The Gallus gallus file we're ingest still exists at NCBI, so it'd improve the metadata slightly to use this URL instead of the DipperCache URL.

Also added some comments in cases where we having to pull data files from DipperCache instead of NCBI


TomConlin:
Note the files are not "missing" from NCBI.  
They never existed which is why they were previously partitioned out locally 
and then migrated to  DipperCache when it came into existance.  

If Gallus gallus is  there/back improving the metadata by referencing it directly
is a good thing.

PR #925, opened 2020-04-22
Fix issue parsing gene ids from interactors in BIOGRID mi.txt file

justaddcoffee:
Some interactors (col1 and col2 of `BIOGRID-ALL-3.5.184.mitab.txt`) are now biogrid CURIEs (e.g. `biogrid:1221377`) instead of gene IDs (e.g. `entrez gene/locuslink:3638344`):

```
(venv) ~/PycharmProjects/dipper *fix_biogrid_id_parsing_issue $ cut -f2 raw/biogrid/BIOGRID-ALL-3.5.184.mitab.txt | grep -v "ID" | grep -v locuslink | head -5
biogrid:1220045
biogrid:1226198
biogrid:1222768
biogrid:1222768
biogrid:1228539
```

This PR changes the code to accept these biogrid CURIEs as gene IDs (and also breaks out the logic for this decision into a method, and tests the method in ` tests/test_biogrid.py`


TomConlin:
looks very good.

aside:
If they internally use the lowercase  curie-prefix  then that is what we would expect to find in the wild then at some point we should consider doing what they do too.

TomConlin:
notes: 
It looks as this ingest has not seen many of the standard refactors most other have
So thanks for fixing the extra stuff you spotted on the way

around line 233; 
the global logging instance for this ingest is 'LOG.'
for  log messages, pep depreciates `.format()` strings in flavor of printf   % syntax.

Does running `make`  as we do to test,  regenerate Julie's alternative curie-prefix  yaml  file? If so please include it in this PR.

On the import typing... there is no point.  shiny always wins.

dipper/sources/BioGrid.py
```diff
@@ -205,6 +207,32 @@ def _get_interactions(self, limit):
 
         return
 
+    def _interactor_to_gene_curie(self, interactor: str) -> Union[str, None]:
```
kshefchek, line 210: Check out Optional - https://docs.python.org/3/library/typing.html#typing.Optional  Same thing but perhaps more readable? it's subjective

justaddcoffee: Okay, done - thanks

justaddcoffee:
> notes:
> It looks as this ingest has not seen many of the standard refactors most other have
> So thanks for fixing the extra stuff you spotted on the way
> 
> around line 233;
> the global logging instance for this ingest is 'LOG.'
> for log messages, pep depreciates .format() strings in flavor of printf % syntax.
> 
> Does running make as we do to test, regenerate Julie's alternative curie-prefix yaml file? If so please include it in this PR.
> 
> On the import typing... there is no point. shiny always wins.


> > Does running make as we do to test, regenerate Julie's alternative curie-prefix yaml file? If so please include it in this PR.

All done, except the make/curie-prefix yaml thing - not sure what this, or how this relates to the BioGrid fix?

TomConlin:
bottom of the Makefile
https://github.com/monarch-initiative/dipper/blob/master/Makefile#L133

Part of Julies identifier work includes collecting various alternative curie-prefixes
so when a non canonical one is used (say in a search)  
it can be translated to a canonical one.

She began with just a manual list 
but we augment it with any we find in the local translation tables.

You added a local translation for an alternative curie-prefix you found in the wild 
`"UNIPROT_PRO": "UniprotKB"`
and it should be automatically added to  the list by running  `make`  then committing any regenerated files that show up under translationtable/generated/
That alternative  might already be there form some previous contact in which case 
nothing would happen. 

just checked :
https://github.com/monarch-initiative/dipper/blob/master/translationtable/generated/prefix_equivalents.yaml

The alternative you found should trigger an update

justaddcoffee:
That `Makefile` no longer runs on my machine, but when I manually run this command:
https://github.com/monarch-initiative/dipper/blob/master/Makefile#L133
no files change.

This command does actually change a file (`prefix_equivalents.yaml`):
https://github.com/monarch-initiative/dipper/blob/master/Makefile#L140
so I've committed this change.

TomConlin:
I would suspect it never really did, but it was never intended to and no one has the resources to address it, so the best I can do is assure you that if you ran on any other dipper developer machine, any monarch server, our Travis or Jenkins instances, it would. 

Thanks for making the effort to generate and commit the file, it is greatly appreciated.

TomConlin:
However the alternative mapping is still not there.

The reason is the discrepancy between what is in `curie_map.yaml`
https://github.com/monarch-initiative/dipper/blob/master/dipper/curie_map.yaml#L231
and what has been added.

justaddcoffee:
Okay, not sure I can help any more with this Biogrid fix. I'm going to close this PR, open it on [Monarch's dipper repo](https://github.com/monarch-initiative/dipper), and you can push whatever fixes are necessary to the YAML files. Does that suit you?

TomConlin:
addressing uniprotkb  curie-prefix is minimal.
leaving  regenerating the prefix file is probably not worth your effort.

justaddcoffee:
> addressing uniprotkb curie-prefix is minimal.
> leaving regenerating the prefix file is probably not worth your effort.

Okay, push whatever changes you think are necessary to this or the other PR (#927)

justaddcoffee:
Handing this off to you @TomConlin

PR #927, opened 2020-04-23
Fix biogrid id parsing issue

justaddcoffee:
Some interactors (col1 and col2 of `BIOGRID-ALL-3.5.184.mitab.txt`) are now biogrid CURIEs (e.g. `biogrid:1221377`) instead of gene IDs (e.g. `entrez gene/locuslink:3638344`):

```
(venv) ~/PycharmProjects/dipper *fix_biogrid_id_parsing_issue $ cut -f2 raw/biogrid/BIOGRID-ALL-3.5.184.mitab.txt | grep -v "ID" | grep -v locuslink | head -5
biogrid:1220045
biogrid:1226198
biogrid:1222768
biogrid:1222768
biogrid:1228539
```

This PR changes the code to accept these biogrid CURIEs as gene IDs (and also breaks out the logic for this decision into a method, and tests the method in ` tests/test_biogrid.py`


justaddcoffee:
@TomConlin push whatever changes you think we need to this PR

PR #930, opened 2020-05-04
HPOA & MGI

TomConlin:
HPOA had a disappearing file I fixed in dippercache (and in my dipper fork) 
over a month ago that somehow missed ever getting pushed to production

there is one error corrected in MGI.py

- Kent found where I had used the var `line` where I should have used `row`.

- in resources/mgi/SQL  there is a whole new set of queries for us to move toward
  - they should have more types of data 
  - they should be less opaque
  - they should be cleaner for us to work with

the queries were written and rewritten several times
the final version is typically much simpler that the previous versions
but I did not like writing them as they are now
because they are several to many times slower to execute. 

the reason to leave them as they are now is they are written to minimize future maintenance if/when MGI's db admins need to change their schema.
and they still try to leave a healthy breadcrumb trail if we need to revisit them.


kshefchek:
+1

PR #945, opened 2020-06-29
HGNC add  default hgnc gene 'in taxon'

TomConlin:
- fix passing empty string as pubmed_id;
 - include new trailing column header 'agr'


balhoff:
Thanks Tom!

PR #949, opened 2020-07-06
set ontobio to 1.18.1

kshefchek:
test to see if this allows us to run RGD, see https://github.com/biolink/ontobio/issues/462


kshefchek:
Looks like this was last working with ontobio 1.14.0

PR #950, opened 2020-07-07
robot and arq pipeline to infer bl categories

kshefchek:
Infer biolink categories in two steps:

1. Generates biolink categories for all classes in monarch.owl with a robot [merge->reason->filter](https://github.com/monarch-initiative/dipper/compare/infer-bl-categories?expand=1#diff-936e5dc6395e8cee532450790933bba5R84) -> [sparql construct](https://github.com/monarch-initiative/dipper/compare/infer-bl-categories?expand=1#diff-cfaf1d78ee710a12258a11b1b4f748fa) pipeline
2. Generates biolink categories for all monarch rdf using Jena ARQ and a [sparql construct query](https://github.com/monarch-initiative/dipper/compare/infer-bl-categories?expand=1#diff-7834201b3a6846a52d021b682a3aae04).

This runs in 38 minutes on monarch4 with make -j 5.  However, it requires 60G of memory for ClinVar, and 30 for robot and some larger turtle files (Panther, MGI).  Since there's no evaluation across multiple triples we could split the ntriples into chunks prior to inferring categories and then merge.

Output from this makefile on our beta dataset is here - https://data.monarchinitiative.org/biolink-model/

This outputs a directory of shim files (just biolink:category triples), and concatenated versions of our dataset (raw output + shim).

I see this as complementary to the hardcoding approach, where we choose inference in the following scenarios:
1. We're already making a rdf:type triple that can be inferred over a rdf:type|subClassOf+ path (stay DRY)
2. We don't know the type at ingest time (CTD and Mesh)
3. There are important subtypes that we don't know at ingest time (Chebi Drug, Metabolite, etc)
4. The type for an identifier is ambiguous and may be applied differently for each source (OMIM percent sign Ids, see coriell)
5. When hardcoding sequence feature categories is non-trivial and error-prone, such as in MGI and ZFIN

Where we should hardcode:
1. Association categories, where inference is possible but messier and would require sparql generation in biolink ml
2. Where we have application driven hacks and want to override category rules (EFO phenotypes in gwascatalog, MPATH and OBA phenotypes in OMIA)
3. Places where we skip rdf:type, such as gene centric ingests where we assume the a more granular gene type will come from another ingest (although we could type these as SO:0000704 and it would be just as well)
4. Any other area where rdf:type|subClassOf|partOf+ entailment is not possible


TomConlin:
+1

scripts/categories/Makefile
```diff
@@ -0,0 +1,144 @@
...
+	$(WGET) $(JENA)
+
+jena/bin/arq: apache-jena-3.15.0.tar.gz jena/
+	tar -C jena/ -xf $<
+
+$(MONARCH_SOURCES): rdf/
+	cd rdf && $(WGET) $(ARCHIVE)/$@
+
+$(QUERIES):
+	sed "s/{DATASET}/$(basename $(notdir $@)).nt/" sparql/infer-categories.sparql.tmpl > $@
```
cmungall, line 119: cool... but maybe we should consider a proper templating framework. What are we doing in GO @balhoff ?

scripts/categories/sparql/bl-categories.sparql
```diff
@@ -0,0 +1,25 @@
...
+
+
+CONSTRUCT {
+  ?type biolink:category ?category .
+}
+WHERE {
+  {
+    GRAPH :biolink-model.ttl  { ?category skos:mappingRelation ?class . }
+    GRAPH :monarch-slim.owl  {
+      ?type (a|rdfs:subClassOf|owl:equivalentClass|owl:sameAs|BFO:0000050)* ?class .
```
cmungall, line 17: You are making the assumption that equiv/sameAs symmetry entailments are asserted

(this is part of a general issue, see https://github.com/TranslatorSRI/reference-kg/issues/9)

I think you also need to account for assertion in the opposite direction with `^`

cmungall (commented):
This is awesome. The symmetry entailment part is the only thing that really needs addressed

scripts/categories/Makefile
```diff
@@ -0,0 +1,144 @@
...
+
+# Monarch release version
+ARCHIVE ?= https://archive.monarchinitiative.org/beta
+
+WGET = /usr/bin/wget --timestamping --no-verbose
+
+JENA = http://archive.apache.org/dist/jena/binaries/apache-jena-3.15.0.tar.gz
+ROBOT_JAR = https://github.com/ontodev/robot/releases/latest/download/robot.jar
+
+BIOLINK_MODEL = https://raw.githubusercontent.com/kshefchek/biolink-model/mappings-for-monarch/biolink-model.ttl
```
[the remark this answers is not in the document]
kshefchek: this is get around this issue - https://github.com/biolink/biolinkml/issues/193

Issue #952, opened 2020-07-16
Get Mondo curies from ClinVar ingest


kshefchek:
ClinVar is starting to use Mondo, eg https://www.ncbi.nlm.nih.gov/clinvar/RCV000005180.3/, however the identifiers are double prefixed

Currently our ingest drops these (ironically) with the log
```RCV000874767 UNKNOWN DISEASE DB:	MONDO:MONDO:0013476```

TomConlin:
elsewhere in dipper we have standardized on keeping only the penultimate prefix
foo.split(':')[-2:]

PR #956, opened 2020-07-21
Clinvar & incidentals

TomConlin:
- Clinvar is starting to use MONDO  (#952)
  - begin recording snapshots of the xml path signature 
- attempting to put zfinslim after zfin in Jenkins
- various reformattings
- regenerated files derived from curie_map that was recently updated
- omia (re?) gets its new url


dipper/sources/ClinVar.py
```diff
@@ -962,26 +962,34 @@ def parse():
                             rcv_disease_id = RCV_TraitXRef.get('ID')
                             break
 
+                    # Accept MONDO if no OMIM or Orphanet  # revisit priority
+                    if rcv_disease_db is None or rcv_disease_id is None:
+                        if rcv_disease_db is not None:
+                            break
+                        for RCV_TraitXRef in RCV_Trait.findall('./XRef[@DB="MONDO"]'):
+                            rcv_disease_db = 'MONDO'  # RCV_TraitXRef.get('DB')
+                            rcv_disease_id = RCV_TraitXRef.get('ID')
```
kshefchek, line 971: IDs already have the MONDO prefix, so I think this will end up double prefixing MONDO:MONDO, RCV_TraitXRef.get('ID')[6:]

TomConlin: line 1071?

kshefchek: ah i missed this, looks good!

PR #959, opened 2020-07-24
InterPro  curie for GO

TomConlin:


cmungall:
I think we need some general principles for adding new CURIE prefixes,
otherwise we are setting ourselves up for clashes

In general I recommend against using a URL that resolves to HTML and is
intended for human. These tend to be unstable. And they also tend to be
different than what others use in RDF, which makes federated
queries/joins/mashups etc harder.

UniProt uses these PURLs:

http://purl.uniprot.org/interpro/IPR000581


If you look in prefix commons, both GO and idot use:

http://identifiers.org/interpro/IPR000581


I would use one of these over a web URL


But let's also sync with biolink, if we are going to represent interpro
URIs in monarch, we want these to be part our datamodel, can you have a go
at adding protein domains to the schema? We have 'gene family' already this
may be a sibling to that

kshefchek:
a simple rule would be anything that resolves to the EBI domain should use identifiers.org, since EBI/Elixir runs id.org there is no risk of attribution issues

cmungall:
Maybe, but there is also an argument that purl.uniprot are the semantic web
experts in the protein space... but either id.org or purl.up are better
than web urls

TomConlin:
There are upstream (external) and downstream (internal) considerations which serve different purposes.
Dipper RDF output at rest can not serve both simultaneously and as we will not be stopping the general public from fetching the data is most appropriate for the data at rest to be in its most general, public friendly form.

Downstream/internally, Monarch/Translator (& other competent entities) can and should use whatever 
name space formats makes the most sense for the identifier fragments in their more refined contexts. 

If rewriting the well defined (in cure_map.yaml) URI is too great a burden on Monarch/Translator processes 
I can (with @mellybelly 's leave) work on making alternative renderings privately available; 
given appropriate curie-prefix to base-uri mapping (yaml) file.

I do not recommend this approach as if these downstream products are going to interface with the wider world
(highly recommended)  they had best be prepared to do this mapping on their own in any case.

cmungall:
I'm not sure I fully understand, but we want to be using the same prefixes
and URIs across projects, anything that serves this aim helps, and reduces
overall churn

PR #960, opened 2020-07-25
basic color legend on graphviz diff files

TomConlin:


TomConlin:
Also put a hyphen in front of some counts in case some folks are more comfortable 
thinking of  `NaN` as zero.

Issue #961, opened 2020-06-08
Diseases which are also Genes


rgb4268:
I have found some instances of diseases, e.g. MONDO:0020119 which have a child which is actually a gene - in this case ZC4H2. There are a few more.

kshefchek:
this is coming from HGNC:24931 being declared an equivalentClass to OMIM:300897, usually HGNC properly xrefs to OMIM genes but this looks like an exception

TomConlin:
We do have the data structures on hand to "trust but verify"  HGNC's assertion that their OMIM dbxref  is in fact to what we also believe is a gene

kshefchek:
false alarm on HGNC, still trying to trace this one

kshefchek:
closer look, this looks like an NCBIGene issue:
<https://www.ncbi.nlm.nih.gov/gene/55906> owl:equivalentClass <http://omim.org/entry/309605>

and this is equivalent to MONDO:0010666

TomConlin:
A bit more context 
omim:309605 has been subsumed into  
omim:314580 which is for WIEACKER-WOLFF SYNDROME  but the original
may have had a different role when it was assigned to  NCBIGene:55906

Dipper does handle the replacement & classifies it as a disease

Also we no longer create an equivalence class as it no longer falls through with `continue`

pnrobinson:
Note also this entry
https://omim.org/entry/301041
I think it is confusing to separate the diseases into XLR and XLD as is being done here

TomConlin:
close w/ #964

PR #964, opened 2020-08-01
NCBI  OMIM disease as gene fix

TomConlin:
- addresses #961
- restore default taxon in more general form. 
   - (no recollection of why I commented them out months ago)
- write out the omim dicts for type & replacement status
- repair incidental test breakage  from category branch
- begin sorting and de-duping ntriples earlier rather than later

### DIFF
--- a/dipper/sources/NCBIGene.py
+++ b/dipper/sources/NCBIGene.py
@@ -145,8 +145,8 @@ class NCBIGene(OMIMSource):
 
         # Defaults
         if self.tax_ids is None:
-            LOG.info('No taxon recived.")  # fallback to defaults')
-            # self.tax_ids = ['9606', '10090', '7955']
+            LOG.info('No taxon recived. Fallback to defaults')
+            self.tax_ids = self.COMMON_TAXON
         else:
             self.tax_ids = [str(x) for x in self.tax_ids]
 
@@ -292,6 +292,7 @@ class NCBIGene(OMIMSource):
 
                 if name != '-':
                     model.addSynonym(gene_id, name)
+
                 if synonyms != '-':
                     for syn in synonyms.split('|'):
                         syn = syn.strip()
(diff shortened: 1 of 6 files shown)


dipper/sources/NCBIGene.py
```diff
@@ -452,13 +453,16 @@ def _add_gene_equivalencies(self, dbxrefs, gene_id, taxon):
...
-                    elif omim_num in self.omim_type and\
+                    if omim_num in self.omim_type and\
                             self.omim_type[omim_num] == self.globaltt['gene']:
                         model.addXref(gene_id, dbxref_curie)
                     else:
-                        continue  # no equivilance between ncbigene and omin-nongene
+                        # OMIM disease/phenotype is not considered a gene at all
+                        # no equivilance between ncbigene and omin-nongene
+                        # and ncbi is never a human clique leader in any case
+                        break
```
kshefchek, line 465: wondering if a break here makes sense, wouldn't this remove any xref after the omim id?

TomConlin: more specifically, any xref after an omim that does not resolve to a gene.  yes

kshefchek: i think these xrefs are still valid, for example for NCBIGene:55906, we previously xref'ed it to ENSEMBL:ENSG00000126970 and OMIM:300897, and made it equivalent to ENSEMBL:ENSG00000126970, OMIM:300897, OMIM:309605, HGNC:24931

On this branch theres only one xref to OMIM:300897 and one eq to OMIM:300897.

Maybe there are cases where a bad OMIM xref indicates the ncbi identifier is not a gene?  But in this case NCBIGene:55906 looks like a gene despite it.

I am in favor of fewer equivalent class triples, especially to ensembl as we tend to get better xrefs from other sources (HGNC, the mod dbs)

TomConlin: Sorry if I am being dense but what are you suggesting?

kshefchek: I think ENSEMBL:ENSG00000126970 should be an xref for NCBIGene:55906

TomConlin: No argument from me, but nothing in this (current) PR prevents it.

TomConlin: Wonder if it is our all screaming caps curie-prefix 'ENSEMBL'

TomConlin: maybe more of a modelling concern but 
isn't being a dbxref redundant considering it is already an equivalenct class which must have uri usable as a xref

kshefchek: looks good to me, I missed where you changed this (was looking at 0deb70)

TomConlin: the UI  presented a sub optimal where it continued to display the old, only flagging it with "updated" in washed out tones.

but all for the best I think as it got me to stare at it harder and fix/streamline that section of logic

Issue #519, opened 2017-09-11
Gene - Variant mixups in mouse data


kshefchek:
On both production and beta we have nodes categorized as both genes and variants for example:

https://scigraph-data.monarchinitiative.org/scigraph/graph/neighbors?id=MGI%3A3811864&depth=1&blankNodes=false&relationshipType=sameAs&direction=BOTH&entail=false&project=*

which comes from mgi.ttl:
http://www.informatics.jax.org/allele/MGI:3811865 owl:sameAs http://www.informatics.jax.org/marker/MGI:3811864

Is this accurate? cc @mbrush

kshefchek:
@cmungall 's suggestion on 8/20/20 data call - remove equivalence and use a GENO relation to link these two entities

PR #971, opened 2020-09-02
Eq and taxon bugs

kshefchek:
Small changes to help with equivalence and category issues:
- NCBI gene: make ensembls xrefs only (no equivalent triples)
- MGI: remove sameAs between marker and alleles (see #519)
- GO: remove interacting taxon as taxon for the gene itself


dipper/sources/GeneOntology.py
```diff
@@ -371,8 +371,10 @@ def process_gaf(self, gaffile, limit, id_map=None):
                         else:
                             model.addSynonym(gene_id, syn)
 
-                for txid in taxon.split('|'):
-                    tax_curie = re.sub(r'taxon:', 'NCBITaxon:', txid)
+                # First taxon is for the gene, after the pipe are interacting taxa
+                tax_curie = taxon.split('|')[0].replace('taxon', 'NCBITaxon')
+                # this is a required field but good to safe
+                if tax_curie:
                     geno.addTaxon(tax_curie, gene_id)
```
TomConlin, line 378: might be interesting to check if exists and if so,  is it different?
also might be unnecessarily slow

kshefchek: if it's different from the expected taxon for the file?

dipper/sources/NCBIGene.py
```diff
@@ -470,7 +473,8 @@ def _add_gene_equivalencies(self, dbxrefs, gene_id, taxon):
                         # and ncbi is never a human clique leader in any case
                         dbxref_curie = None
                         continue
-                # designate clique leaders
+
+                # designate clique leaders and equivalentClass/sameAs triples
```
TomConlin, line 477: based on  dict dipper/resources/clique_leaders.yaml

TomConlin:
+1

TomConlin (approved):
detecting discrepancies within a file  which most likely is not worth looking for unless we notice is being an actual problem

Issue #972, opened 2020-09-04
Disease-gene mixups for biolink:categories in OMIM ingest


kshefchek:
@deepakunni3 is reporting that the KGX ingest found 161 OMIM ids that are both diseases and genes:
[errant-omim.txt](https://github.com/monarch-initiative/dipper/files/5172205/errant-omim.txt)

TomConlin:
a bit more including which ingests say what 
[errant_blcat.txt](https://github.com/monarch-initiative/dipper/files/5172709/errant_blcat.txt)

kshefchek:
for the september data I've removed the bl categories that come from dipper in the ntriples file, but they can be fetched from the inferred files - https://archive.monarchinitiative.org/202009/rdf/blshim/omim.nt or https://archive.monarchinitiative.org/202009/rdf/blcategories/omim.nt

kshefchek:
fixed with https://github.com/monarch-initiative/dipper/pull/973

PR #973, opened 2020-09-04
remove category from G2D

kshefchek:
See #972, we should assign categories based on the omim type, not based on its position in a gene to disease association (which sometimes includes non diseases)


TomConlin:
+1

PR #979, opened 2020-09-22
Disambiguate dcat:Distribution from dcat:distribution

nicholsn:
Currently the dataset descriptions created by `dipper` only have the class IRI for `dcat:Distribution` and not the property IRI of `dcat:distribution`.

This PR updates the `GLOBAL_TERMS.yaml` to include the property as part of the map and updates the code to use the property correctly.

Fun fact, I detected this using the `kgx` package while validating the results after running dipper-etl.py for ensembl. Here is an example of the error and dataset descriptor that isn't valid.

```
[ERROR][INVALID_EDGE_LABEL] https://archive.monarchinitiative.org/20200908/#ensembl-https://archive.monarchinitiative.org/20200908/rdf/ensembl.ttl - Edge label 'DistributionLevel' is not in snake_case form
```

```
@prefix biolink: <https://w3id.org/biolink/vocab/> .
@prefix dcat: <http://www.w3.org/ns/dcat#> .
@prefix dcterms: <http://purl.org/dc/terms/> .
@prefix dctypes: <http://purl.org/dc/dcmitype/> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .
@prefix pav: <http://purl.org/pav/> .
@prefix schema: <http://schema.org/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

<http://www.ensembl.org/biomart/martservice?> pav:retrievedOn "2020-09-08"^^xsd:date .

<https://archive.monarchinitiative.org/#ensembl> a dctypes:Dataset,
        owl:Ontology ;
    dcterms:Publisher <https://monarchinitiative.org/> ;
    dcterms:identifier "MonarchArchive:#ensembl" ;
    dcterms:source <http://uswest.ensembl.org> ;
    dcterms:title "ENSEMBL" ;
    schema:logo <https://github.com/monarch-initiative/monarch-ui/blob/master/public/img/sources/source-ensembl.png> ;
    owl:versionIRI <https://archive.monarchinitiative.org/20200908/#ensembl> .

<https://archive.monarchinitiative.org/20200908/#ensembl> a dctypes:Dataset ;
    dcterms:Publisher <https://monarchinitiative.org/> ;
    dcterms:created "2020-09-08"^^xsd:date ;
    dcterms:creator <https://monarchinitiative.org/> ;
    dcterms:isVersionOf <https://archive.monarchinitiative.org/#ensembl> ;
    dcterms:source <http://www.ensembl.org/biomart/martservice?> ;
    dcterms:title "ENSEMBL Monarch version 20200908" ;
    pav:version "2020-09-08"^^xsd:date ;
    dcat:Distribution <https://archive.monarchinitiative.org/20200908/rdf/ensembl.ttl> ;
    biolink:category biolink:DataSetVersion .

<https://archive.monarchinitiative.org/20200908/rdf/ensembl.ttl> a dctypes:Dataset,
        dcat:Distribution ;
    dcterms:Publisher <https://monarchinitiative.org/> ;
    dcterms:created "2020-09-08"^^xsd:date ;
    dcterms:creator <https://monarchinitiative.org/> ;
    dcterms:downloadURL <https://archive.monarchinitiative.org/20200908/rdf/ensembl.ttl> ;
    dcterms:format <https://www.w3.org/TR/turtle/> ;
    dcterms:license <https://project-open-data.cio.gov/unknown-license/> ;
    dcterms:title "ENSEMBL distribution ttl" ;
    pav:createdWith <https://github.com/monarch-initiative/dipper> ;
    pav:version "2020-09-08"^^xsd:date .
```

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -100,7 +100,8 @@ class DatasetTestCase(unittest.TestCase):
         cls.iri_retrieved_on = URIRef(cls.base_pav + "retrievedOn")
         cls.iri_creator = URIRef(cls.base_dcterms + "creator")
         cls.iri_is_version_of = URIRef(cls.base_dcterms + "isVersionOf")
-        cls.iri_distribution = URIRef(cls.base_dcat + "Distribution")
+        cls.iri_distribution = URIRef(cls.base_dcat + "distribution")
+        cls.iri_Distribution = URIRef(cls.base_dcat + "Distribution")
         cls.iri_created_with = URIRef(cls.base_pav + "createdWith")
         cls.iri_format = URIRef(cls.base_dcterms + "format")
         cls.iri_download_url = URIRef(cls.base_dcterms + "downloadURL")
@@ -342,7 +343,7 @@ class DatasetTestCase(unittest.TestCase):
         triples = list(self.dataset.graph.triples(
             (self.distribution_level_IRI_ttl,
              self.iri_rdf_type,
-             self.iri_distribution)))
+             self.iri_Distribution)))
         self.assertTrue(len(triples) == 1,
                         "missing version level type distribution triple")
 
--- a/translationtable/GLOBAL_TERMS.yaml
+++ b/translationtable/GLOBAL_TERMS.yaml
@@ -58,7 +58,7 @@
 "Microcell hybrid": "CLO:0036939",
 "Chorionic villus-derived cell line": "CLO:0036940",
 # "Data Catalog Vocabulary": "dcat:!!!"
-"distribution": "dcat:Distribution",
+"distribution": "dcat:distribution",
 "Distribution": "dcat:Distribution",
 # "dublin core terms": "dcterms:!!!",
 "Date Created": "dcterms:created",
--- a/translationtable/GLOBAL_TERMS.yaml
+++ b/translationtable/GLOBAL_TERMS.yaml
@@ -59,6 +59,7 @@
 "Chorionic villus-derived cell line": "CLO:0036940",
 # "Data Catalog Vocabulary": "dcat:!!!"
 "distribution": "dcat:Distribution",
+"Distribution": "dcat:Distribution",
 # "dublin core terms": "dcterms:!!!",
 "Date Created": "dcterms:created",
 "description": "dcterms:description",
--- a/dipper/models/Dataset.py
+++ b/dipper/models/Dataset.py
@@ -290,7 +290,7 @@ class Dataset:
         self.model.addType(self.distribution_level_turtle_curie,
                            self.globaltt['Dataset'])
         self.model.addType(self.distribution_level_turtle_curie,
-                           self.globaltt['distribution'])
+                           self.globaltt['Distribution'])
         self.graph.addTriple(self.distribution_level_turtle_curie,
                              self.globaltt['title'],
                              self.ingest_title +


translationtable/GLOBAL_TERMS.yaml
```diff
@@ -58,7 +58,8 @@
 "Microcell hybrid": "CLO:0036939",
 "Chorionic villus-derived cell line": "CLO:0036940",
 # "Data Catalog Vocabulary": "dcat:!!!"
-"distribution": "dcat:Distribution",
+"distribution": "dcat:distribution",
```
TomConlin, line 61: Thanks @nicholsn 

We would not want both.
The key is intended to be (move towards) the term's one proper label,

We endevor to keep the dict invertable to find labels from terms as well.

tests/test_dataset.py
```diff
@@ -100,7 +100,8 @@ def setUpClass(cls):
         cls.iri_retrieved_on = URIRef(cls.base_pav + "retrievedOn")
         cls.iri_creator = URIRef(cls.base_dcterms + "creator")
         cls.iri_is_version_of = URIRef(cls.base_dcterms + "isVersionOf")
-        cls.iri_distribution = URIRef(cls.base_dcat + "Distribution")
+        cls.iri_distribution = URIRef(cls.base_dcat + "distribution")
```
TomConlin, line 103: please omit the lowercase here as well

translationtable/GLOBAL_TERMS.yaml
nicholsn: Thanks for the response, @TomConlin! 

I was coming at this from the perspective that the generated dataset turtle documents produced by dipper use `dcat:Distribution` as both an object property and a type, whereas these URIs actual have different meanings in the dcat standard:

> [dcat:Distribution](https://www.w3.org/TR/vocab-dcat-2/#Class:Distribution): A specific representation of a dataset. A dataset might be available in multiple serializations that may differ in various ways, including natural language, media-type or format, schematic organization, temporal and spatial resolution, level of detail or profiles (which might specify any or all of the above).

> [dcat:distribution](https://www.w3.org/TR/vocab-dcat-2/#Property:dataset_distribution): An available distribution of the dataset.

Can you help me better understand what you mean by "the term's one proper label"? How do you handle having the correct RDF serialization in this case?

TomConlin: the curie-prefix  `dcat:` has a single base-uri (namespace) mapping here:

https://github.com/monarch-initiative/dipper/blob/master/dipper/curie_map.yaml#L25

which is as they recommend 
`http://www.w3.org/ns/dcat#`

yielding
https://www.w3.org/ns/dcat#Distribution
and
https://www.w3.org/ns/dcat#distribution

Both resolving to the same page 
with no directly relevant information....

I guess it will work out.
I had thought you were accommodating
different representations of the same term.. 

but I still  find relying on terms differing only by 
a single letter case 
to be unfortunate and expect it will be fragile.

nicholsn: Good point... with dcat 1.0 the redirect would take you to the #ref of the specific term in the spec... looks like with v2 they don't have the redirect, but at least content negotiation is working `curl -H 'Accept:text/turtle' https://www.w3.org/ns/dcat#distribution`

And I agree the whole CamelCase for classes and snakeCase for properties can be brittle compared to obo-style accession ids.

Anyhow, thanks for the merge! Really great tool and datasets that you all are making available. Hoping to learn more about the whole biolink\monarch ecosystem

TomConlin: Thank you for the PR!

PR #983, opened 2020-10-10
Adendum to PR #982   (Digest bnode ids & refactor)

TomConlin:
bnode ids digested within the model classes needed an alternative 
to the digest function in Source.make_id()'  unless we change all the signatures to pass it.

zfin & mgi  run locally , expect others will as well


TomConlin:
unrelated? Ensembl BioMart failure 
(I can't get to OLS  now either)

PR #989, opened 2020-10-22
AnimalQTL updates

TomConlin:
they moved to newer assemblies 
basic refactors

include some wormbase artifacts for reference 
to help next time they change something


TomConlin:
it has been an hour and travis is still stuck in its queue to start

PR #991, opened 2020-10-23
change dipper cache host

kshefchek:
Proposing this until we have proper hosts/dns routing


TomConlin:
Unfortunately it eliminates the ability to test from off site which remains important to me. 

I would prefer a solution that does not require all testing to originate from  behind cgrb's gateway.

TomConlin:
We could parameterize the cache location and it not be a bad thing...

kshefchek:
going to close this for now since the network issue seems transient

PR #995, opened 2020-11-03
update common taxa in jenkins

kshefchek:
although I think this will break parsers where the taxon is not yet configured, should we be defensive and fail hard or output a warning?


TomConlin:
Sure, it is a great day to blow up all the things.
Plus prod is freshly updates so we have a bit of grace 
to fix or back out. 
I think I left any species out of the common set that could not work
when the common set was used, and this is most likely a departure from that, but one that if accommodated would make including future species easier

TomConlin:
string unit test needing more is not suprising

kshefchek:
i'm going to close as i'm pretty sure this is going to break something when a taxon is not supported by a source

PR #1000, opened 2020-11-25
Fix WormBase prefix URI to Identifiers.org entry

RichardBruskiewich:


TomConlin:
Unless Wormbase changes their identifiers to a third-party url re-writer
dipper will continue to use the identifier-minter's url for its public repository
and rely on downstream processes to rewrite their own strings as required.

PR #1002, opened 2020-11-27
Split PATO namespaces

RichardBruskiewich:
Need two distinct PATO namespaces, one for regular terms, the other for PATO properties. See @balhoff observations in https://github.com/biolink/biolink-model/issues/516


RichardBruskiewich:
Cancelling this PR after chatting with @TomConlin about where to host the change required.

PR #1003, opened 2020-12-09
Skip normal phenotypes in MGI

kshefchek:
Skip over phenotypes marked as normal in MGI, since we do not yet model normal phenotypes or negation


TomConlin:
+1

PR #1004, opened 2020-12-16
add new variant type 'haplotype, single variant'

kshefchek:
clinvar has a new variant type called 'haplotype, single variant', translating it to our global term for 'haplotype'


TomConlin:
+1

